PR #1, opened 2021-11-30
DM-31723: Update code for new name and location

timj:
Prior to this the code assumed that it was in daf_butler and had no build infrastructure in the history.


TallJimbo (approved):

.github/workflows/black.yaml
```diff
@@ -0,0 +1,17 @@
...
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
+        with:
+          options: "--check --verbose --diff --line-length 110 --target-version py38"
```
timj, line 16: @TallJimbo I've turned on black in this repo to see how we get on. These options seemed reasonable.

timj: and I realized I could put those options in the `pyproject.toml` file.

TallJimbo: We probably want `isort` in the same action; here's what I did recently for the combo:

https://github.com/lsst-dm/cell_coadds/blob/main/.github/workflows/formatting.yaml
https://github.com/lsst-dm/cell_coadds/blob/main/pyproject.toml

timj: Okay done. I've also committed a pre-commit config file.

PR #4, opened 2022-01-24
DM-33360: Improve performance of https as_local() method

airnandez:
## Checklist

- [ ] ran Jenkins

Unit tests executed in my development environment passed. Could not run Jenkins.

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -222,7 +222,7 @@ def finalurl(r: requests.Response) -> str:
 
 
 # Tuple (path, block_size) pointing to the location of a local directory
-# to save temporary files and its associated block size
+# to save temporary files and the block size of the underlying file system
 _TMPDIR = None
 
 
@@ -238,7 +238,7 @@ def _get_temp_dir() -> Tuple[str, int]:
     # Use the value of environment variables 'LSST_BUTLER_TMPDIR' or
     # 'TMPDIR', if defined. Otherwise use current working directory
     tmpdir = os.getcwd()
-    for dir in (os.getenv(v) for v in ('LSST_BUTLER_TMPDIR', 'TMPDIR')):
+    for dir in (os.getenv(v) for v in ("LSST_BUTLER_TMPDIR", "TMPDIR")):
         if dir and os.path.isdir(dir):
             tmpdir = dir
             break
@@ -353,8 +353,9 @@ class HttpResourcePath(ResourcePath):
             suffix=self.getExtension(), buffering=buffering, dir=tmpdir, delete=False
         ) as tmpFile:
             with time_this(
-                log, msg="Downloading %s [length=%s] to local file %s [chunk_size=%d]",
-                args=(self, r.headers['Content-Length'], tmpFile.name, buffering)
+                log,
+                msg="Downloading %s [length=%s] to local file %s [chunk_size=%d]",
+                args=(self, r.headers.get("Content-Length"), tmpFile.name, buffering),
             ):
                 for chunk in r.iter_content(chunk_size=buffering):
                     tmpFile.write(chunk)
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -227,9 +227,9 @@ _TMPDIR = None
 
 
 def _get_temp_dir() -> Tuple[str, int]:
-    """Returns the temporary directory path and block size.
+    """Return the temporary directory path and block size.
 
-    Caches its results in _TMPDIR.
+    This function caches its results in _TMPDIR.
     """
     global _TMPDIR
     if _TMPDIR:
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -25,7 +25,7 @@ from typing import TYPE_CHECKING, Optional, Tuple, Union
 
 from lsst.utils.timer import time_this
 from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.util.retry import Retry
+from urllib3.util.retry import Retry
 
 from ._resourcePath import ResourcePath
 
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -238,7 +238,7 @@ def _get_temp_dir() -> Tuple[str, int]:
     # Use the value of environment variables 'LSST_BUTLER_TMPDIR' or
     # 'TMPDIR', if defined. Otherwise use current working directory
     tmpdir = os.getcwd()
-    for dir in (os.getenv(v) for v in ("LSST_BUTLER_TMPDIR", "TMPDIR")):
+    for dir in (os.getenv(v) for v in ("LSST_RESOURCES_TMPDIR", "TMPDIR")):
         if dir and os.path.isdir(dir):
             tmpdir = dir
             break
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -221,6 +221,37 @@ def finalurl(r: requests.Response) -> str:
     return destination_url
 
 
+# Tuple (path, block_size) pointing to the location of a local directory
+# to save temporary files and its associated block size
+_TMPDIR = None
+
+
+def _get_temp_dir() -> Tuple[str, int]:
+    """Returns the temporary directory path and block size.
+
+    Caches its results in _TMPDIR.
+    """
+    global _TMPDIR
+    if _TMPDIR:
+        return _TMPDIR
+
+    # Use the value of environment variables 'LSST_BUTLER_TMPDIR' or
+    # 'TMPDIR', if defined. Otherwise use current working directory
+    tmpdir = os.getcwd()
+    for dir in (os.getenv(v) for v in ('LSST_BUTLER_TMPDIR', 'TMPDIR')):
+        if dir and os.path.isdir(dir):
+            tmpdir = dir
+            break
+
+    # Compute the block size as 256 blocks of typical size
+    # (i.e. 4096 bytes) or 10 times the file system block size,
+    # whichever is higher. This is a reasonable compromise between
+    # using memory for buffering and the number of system calls
+    # issued to read from or write to temporary files
+    fsstats = os.statvfs(tmpdir)
+    return (_TMPDIR := (tmpdir, max(10 * fsstats.f_bsize, 256 * 4096)))
+
+
 class HttpResourcePath(ResourcePath):
     """General HTTP(S) resource."""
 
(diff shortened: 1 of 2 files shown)


python/lsst/resources/http.py
```diff
@@ -313,13 +313,35 @@ def _as_local(self) -> Tuple[str, bool]:
         temporary : `bool`
             Always returns `True`. This is always a temporary file.
         """
-        log.debug("Downloading remote resource as local file: %s", self.geturl())
+        log.debug(f"Downloading {self.geturl()} as local file")
```
timj, line 316: We are not supposed to use f-strings in any log messages since it is deemed to be more efficient to defer the string interpolation until we know the log message will be issued. I think we can probably delete this log message completely since the `time_this` code below is already going to issue a log message.

python/lsst/resources/http.py
```diff
@@ -313,13 +313,35 @@ def _as_local(self) -> Tuple[str, bool]:
...
             raise FileNotFoundError(f"Unable to download resource {self}; status code: {r.status_code}")
-        with tempfile.NamedTemporaryFile(suffix=self.getExtension(), delete=False) as tmpFile:
+
+        # Compute the block size to the file system where temporary files are
+        # to be written. Use the locations pointed to by the environment
+        # variables used by tempfile.mkstemp() to determine the paths of the
+        # temporary directories.
+        buffering = -1
+        chunk_size = None
+        for path in (os.getenv(var) for var in ('TMPDIR', 'TEMP', 'TMP')):
```
timj, line 327: Why are we guessing where the file is going to be written? We know where it is going to be written.

python/lsst/resources/http.py
```diff
@@ -313,13 +313,35 @@ def _as_local(self) -> Tuple[str, bool]:
...
+                # whichever is higher. This is a reasonable compromise between
+                # using memory for buffering and the number of system calls
+                # issued to read from the underlying socket and to write to
+                # the temporary file
+                fsstats = os.statvfs(path)
+                buffering = chunk_size = max(10*fsstats.f_bsize, 256*4096)
+                break
+
+        with tempfile.NamedTemporaryFile(suffix=self.getExtension(), buffering=buffering,
+                                         delete=False) as tmpFile:
```
timj, line 340: Why can't we do the chunk calculation in this block?
```python
fsstats = os.statvfs(tmpFile.name)
```

python/lsst/resources/http.py
```diff
@@ -313,13 +313,35 @@ def _as_local(self) -> Tuple[str, bool]:
...
+            log.debug(f"Downloading {int(r.headers['Content-Length'])} bytes from {self.geturl()} "
```
timj, line 341: Can you embed this content into the `time_this` call below so we get one informative debug log message for the transfer?

timj (changes requested):
You need to run `black` on the file to make it pass the new formatting requirements. I'm a bit confused by the mypy complaint about a missing stub -- I do install the requests type stubs.

My main comment is I don't know why we can't work out the block size for the temp file we did get rather than guess in advance where the temp file will end up.

I'm intrigued that it's more important to get the local file system block size correct than any kind of network transfer chunk size.

timj:
Yes, when I relocated the code from daf_butler I lost some test coverage -- specifically there is not a single test of `as_local` because in butler that was being tested explicitly in the datastore. I will need to clean up the tests in this package.

python/lsst/resources/http.py
airnandez: > My main comment is I don't know why we can't work out the block size for the temp file we did get rather than guess in advance where the temp file will end up.

I think it is useful to have the same size of buffer for both reading from the socket and writing to the temporary file. The size of the temporary file buffer is passed to at the creation of the file (ie. the call to `tempfile.NamedTemporaryFile(...)`), this is why it is computed beforehand. The same size is used to read data from the network.

> I'm intrigued that it's more important to get the local file system block size correct than any kind of network transfer chunk size.

I think both are important to reduce system calls: if I read from the network chunks of 4 KB and write to the temporary file chunks of 1 MB, I will need much more `read` system calls than `write` system calls.

python/lsst/resources/http.py
airnandez: > Why are we guessing where the file is going to be written? We know where it is going to be written.

These are the locations where `tempfile.mkstemp()` creates the temporary file, if any of them are defined. At this moment in the execution, it is not known which one will be selected, so we have to inspect them in order.

python/lsst/resources/http.py
airnandez: It is possible, but then we could get an asymmetry between the socket buffer size and the temporary file buffer size.

python/lsst/resources/http.py
airnandez: I did think of that but thought that the log message with all the details of the chunk size and the remote contents size made sense in debug mode only. If I embed that message in `time_this` will it be systematically shown even in other log levels?

python/lsst/resources/http.py
airnandez: > We are not supposed to use f-strings in any log messages...

Understood. I didn't know that.

python/lsst/resources/http.py
timj: `time_this` can be told what level to use for it's message. It defaults to `logging.DEBUG`.

airnandez:
> I also forgot to say that I'm concerned that we don't have any of this in the test coverage -- that's probably my fault rather than anything else.

To be clear, in my local development environment I did `scons -j 16` and made sure there were no failing tests. I hope this is what I was expected to do.

python/lsst/resources/http.py
timj: I had missed the fact that the buffering parameter is passed to `NamedTemporaryFile`. 😦

timj:
Jenkins should work fine with this since the branch is not a fork. The main issue is that we don't actually test `as_local` at all. Maybe you can add a quick test like:

```diff
diff --git a/tests/test_uri.py b/tests/test_uri.py
index ee6eca3..bc016df 100644
--- a/tests/test_uri.py
+++ b/tests/test_uri.py
@@ -938,6 +938,11 @@ class WebdavURITestCase(unittest.TestCase):
         with self.assertRaises(FileNotFoundError):
             self.notExistingFileResourcePath.read()
 
+        with self.existingFileResourcePath.as_local() as local_uri:
+            self.assertTrue(local_uri.isLocal)
+            content = local_uri.read().decode()
+            self.assertEqual(content, "It works!")
+
     @responses.activate
     def testWrite(self):
 
```

python/lsst/resources/http.py
airnandez: I just checked and confirm that `time_this` only shows the message when the log level is `DEBUG`.

timj: The main difference is that the timer log message has a `timer.` prefix by default -- I doubt that is going to be an issue but if it is an issue maybe you should keep the log output separate.

python/lsst/resources/http.py
airnandez: This is an excerpt of the documentation of [tempfile.mkstemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp)

> _If dir is not None, the file will be created in that directory; otherwise, a default directory is used. The default directory is chosen from a platform-dependent list, but the user of the application can control the directory location by setting the TMPDIR, TEMP or TMP environment variables._ 

Should we then enforce, at the application level, that the directory for temporary file creation must be pointed to `TMPDIR` (which is a POSIX-defined variable) the dfault value being `/tmp` ? I find that a cleaner solution than to create every temporary file twice just to retrieve the block size and could even retrieve that block size once and save it.

timj: If we believe that the block size will be fixed for the entire process (modulo shenanigans of tests redefining TMPDIR) then can't we add a class method for doing this where we will create a throwaway temp file once and cache the result?

I suppose another approach is to use the `dir` parameter and define exactly where we want the file to end up -- we can use `tmpdir = os.environ.get("TMPDIR", ResourcePath(".", forceAbsolute=True, forceDirectory=True))` (or do the loop you do now with `cwd` at the end)  and then examine `tmpdir` directly. Still might be worth caching this rather than doing the fsstat call every time.

```python
_TMPDIR = None
def _get_temp_dir():
    if _TMPDIR is None:
        for tmpdir in (...):
            ...
        _TMPDIR = (tmpdir, buffering)
    return _TMPDIR

def _as_local(self):
    tmpdir, buffering = _get_temp_dir()
    ...
```
or something?

airnandez:
I implemented the agreed functionality and added the suggested test to `test_uri.py`. I added the possibility of defining the environment variable `LSST_BUTLER_TMPDIR` to specify the location to use for downloading temporary files, in case the user wants that to be different from the standard `TMPDIR`, which is also considered if the former is not defined. If none of them are defined, the current working directory will be used as temporary directory.

I don't quite understand the errors of the `mypy  / lint` check. As far as I can see, the module `requests.packages.urllib3.util.retry` is actually imported by `http.py`

python/lsst/resources/http.py
```diff
@@ -221,6 +221,37 @@ def finalurl(r: requests.Response) -> str:
...
+
+
+def _get_temp_dir() -> Tuple[str, int]:
+    """Return the temporary directory path and block size.
+
+    This function caches its results in _TMPDIR.
+    """
+    global _TMPDIR
+    if _TMPDIR:
+        return _TMPDIR
```
timj, line 236: It might be worth adding a second `as_local()` test to ensure that we get code coverage on this line.

python/lsst/resources/http.py
```diff
@@ -221,6 +221,37 @@ def finalurl(r: requests.Response) -> str:
...
+
+    # Use the value of environment variables 'LSST_BUTLER_TMPDIR' or
+    # 'TMPDIR', if defined. Otherwise use current working directory
+    tmpdir = os.getcwd()
+    for dir in (os.getenv(v) for v in ("LSST_BUTLER_TMPDIR", "TMPDIR")):
+        if dir and os.path.isdir(dir):
+            tmpdir = dir
```
timj, line 243: The implication from the code coverage is that neither of these environment variables are set. What you can do is in the test do something like:

```python
with unittest.mock.patch.dict(os.environ, {"TMPDIR": TESTDIR}):
    with self.existingFileResourcePath.as_local() as local_uri:
        ...
        self.assertIsNotNone(local_uri.relative_to(ResourcePath(TESTDIR)))
```
which would check that the local URI is a subdirectory of TESTDIR.

python/lsst/resources/http.py
```diff
@@ -221,6 +221,37 @@ def finalurl(r: requests.Response) -> str:
...
+    This function caches its results in _TMPDIR.
+    """
+    global _TMPDIR
+    if _TMPDIR:
+        return _TMPDIR
+
+    # Use the value of environment variables 'LSST_BUTLER_TMPDIR' or
+    # 'TMPDIR', if defined. Otherwise use current working directory
+    tmpdir = os.getcwd()
+    for dir in (os.getenv(v) for v in ("LSST_BUTLER_TMPDIR", "TMPDIR")):
```
timj, line 241: This package is nothing to do with Butler so it's not an obvious name. I know that `http.py` has plenty of other environment variables that do but I didn't change them because I was worried I would break your environment. I think this should probably be `LSST_RESOURCES_TMPDIR` (I do use a `LSST_RESOURCES_TEST_TMP` in tests but you wouldn't use that).

timj (approved):
I'm happy with this. My main comments concern test coverage.

timj:
@airnandez I fixed the mypy problem and pushed a fix.

python/lsst/resources/http.py
timj: It occurs to me that this is not going to be reliable because we cache the `_TEMPDIR` and so by definition are not allowing the answer to change between calls. I guess the test could force the `_TEMPDIR` global to be `None` before re-running the test (it's allowed to know about the implementation).

timj:
@airnandez I also added a couple of extra tests so we get full coverage of the patch. I'm happy with this now. I'll probably have to fix the S3 code path now to make it match the `LSST_RESOURCES_TMPDIR` usage. Please take a look and I can merge this if you are happy.

airnandez:
> I know that http.py has plenty of other environment variables that do but I didn't change them because I was worried I would break your environment.

That is the reason I choose the `LSST_BUTLER_...` prefix, but you are right that the `resources` module is not specific to butler, so that name is not appropriate.

Changing the other `LSST_BUTLER_*`  environment variables would break things for us but I think it is better to break them now and adapt than to live with that forever. I propose we address those changes in a separate pull request I would be happy to contribute, if you agree.

airnandez:
Thanks a lot for all these final touches (and for your patience!). I am happy with you merging these changes.

timj:
> I propose we address those changes in a separate pull request I would be happy to contribute, if you agree.

Works for me.

timj:
I forgot to answer this question:

> To be clear, in my local development environment I did scons -j 16 and made sure there were no failing tests. I hope this is what I was expected to do.

Yes, but my comment referred to the coverage report on this pull request where it showed which lines were not covered by tests.

PR #6, opened 2022-02-11
DM-33637: Use upload_file and add progress reporting for S3

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/s3.py
+++ b/python/lsst/resources/s3.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import logging
 import re
 import tempfile
+import threading
 
 __all__ = ("S3ResourcePath",)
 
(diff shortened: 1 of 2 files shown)


TallJimbo (approved):
Looks good.

Makes me wonder if we should the progress stuff here and in daf_butler up into utils so they could both be more broadly used.  I don't see a lot of logic duplication, but it'd be nice to have a consistent interface for progress reporting with different displays (e.g. bars vs. logs).  But I think that's another ticket, and only if you're enthusiastic; I don't think I can afford to take it on any time soon).

timj:
Yes, we never really had a reason to move your progress bar code out of butler but this does seem like it might well be a reason to do that.

PR #7, opened 2022-02-21
DM-33769: Improve timeout, retry and connection persistence of webDAV client

airnandez:
## Checklist

- [x] ran Jenkins


python/lsst/resources/http.py
```diff
@@ -188,7 +188,7 @@ def _send_expect_header_on_put() -> bool:
 
 
 @functools.lru_cache
-def _is_webdav_endpoint(path: Union[ResourcePath, str]) -> bool:
+def isWebdavEndpoint(path: Union[ResourcePath, str]) -> bool:
```
timj, line 191: This needs some though. I don't like it being a public API at all and it's only public because one test in butler uses it and I'm not sure if that test really should be doing so.

airnandez: The class `HttpResourcePath` exposes the (property) method `is_webdav_endpoint`. Maybe the test in butler can use that method. If that is the case, I would vote for the function `isWebdavEndpoint` not to be exposed.

timj: No. The user doesn't know what type of scheme they have so there should be no public properties that are specific to a particular scheme (so really it should be `_is_webdav_endpoint`). Code should not have to change for an S3 scheme for example.

timj: Maybe we call it `_is_webdav_endpoint` and use it in the butler test as a clue that it's a private attribute that can only be used in special scenarios? Then we'd still be noting that it's not a general ResourcePath feature whilst allowing us not to have to export the helper function

airnandez: > Maybe we call it `_is_webdav_endpoint` and use it in the butler test
That seems a reasonable compromise. Butler tests would need modification then.

Please confirm I should proceed renaming `isWebdavEndpoint` by `_is_webdav_endpoint`.

timj: Rename is fine. It's a minor patch of daf_butler (which would have to be done on the same ticket).

airnandez: Modification done and pull request for `daf_butler` [submitted](https://github.com/lsst/daf_butler/pull/652) (ticket DM-33769)

airnandez:
Any further comments welcome.

python/lsst/resources/http.py
```diff
@@ -252,26 +247,79 @@ def _get_temp_dir() -> Tuple[str, int]:
...
+    @property
+    def upload_session(self) -> requests.Session:
+        """Client object to address remote resource for PUT requests."""
+        cls = type(self)
+        if cls._upload_session:
+            return cls._upload_session
+
+        log.debug("Creating new HTTP session for PUT requests: %s", self.geturl())
+        cls._upload_session = _get_http_session(self)
+        return cls._upload_session
```
timj, line 322: Similar comment above regarding using a class attribute as an instance property when the root URI can change.

timj (approved):
Thanks for doing this. I have one big concern about the caching of `cls._session`.

python/lsst/resources/http.py
```diff
@@ -34,139 +37,159 @@
...
+
+
+def _get_http_session(path: ResourcePath, persist: bool = True) -> requests.Session:
     """Create a requests.Session pre-configured with environment variable data.
 
+    Parameters
+    ----------
+    path : `ResourcePath`
+        URL to a resource in the remote server for which the session is to be
+        created
```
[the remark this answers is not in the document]
airnandez: Fixed.

python/lsst/resources/http.py
```diff
@@ -34,139 +37,159 @@
...
+    log.warning(
+        "Neither LSST_HTTP_AUTH_BEARER_TOKEN nor (LSST_HTTP_AUTH_CLIENT_CERT and LSST_HTTP_AUTH_CLIENT_KEY)"
+        " are initialized. No client authentication enabled."
+    )
+    return session
 
 
-def isTokenAuth() -> bool:
-    """Return the status of bearer-token authentication.
+@functools.lru_cache
```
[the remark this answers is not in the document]
airnandez: Fixed.

tests/test_http.py
```diff
@@ -156,6 +152,11 @@ def test_remove(self):
         with self.assertRaises(FileNotFoundError):
             self.notExistingFileResourcePath.remove()
 
+        with self.assertRaises(FileNotFoundError):
```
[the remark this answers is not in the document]
airnandez: Fixed.

tests/test_http.py
```diff
@@ -236,6 +242,132 @@ def test_parent(self):
...
+        with unittest.mock.patch.dict(
+            os.environ,
+            {"LSST_HTTP_AUTH_CLIENT_CERT": client_cert, "LSST_HTTP_AUTH_CLIENT_KEY": client_key},
+            clear=True,
+        ):
+            # Ensure the session client certificate is initialized when
+            # only the owner can read the private key file
+            os.chmod(client_key, stat.S_IRUSR)
+            session = _get_http_session(self.baseURL)
+            self.assertTrue(session.cert[0] == client_cert)
```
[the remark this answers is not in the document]
airnandez: Fixed.

tests/test_http.py
```diff
@@ -236,6 +242,132 @@ def test_parent(self):
...
+            # Ensure an exception is raised if either group or other can access
+            # the private key file
+            for mode in (stat.S_IRGRP, stat.S_IWGRP, stat.S_IXGRP, stat.S_IROTH, stat.S_IWOTH, stat.S_IXOTH):
+                os.chmod(client_key, stat.S_IRUSR | mode)
+                with self.assertRaises(PermissionError):
+                    _get_http_session(self.baseURL)
+
+    def test_sessions(self):
+        path = ResourcePath(self.baseURL)
+        self.assertTrue(self.baseURL.session is not None and self.baseURL.session == path.session)
```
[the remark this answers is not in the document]
airnandez: Fixed.

tests/test_webdavutils.py
```diff
@@ -19,12 +19,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import os
 import unittest
 
 import requests
 import responses
-from lsst.resources.http import finalurl, isTokenAuth, isWebdavEndpoint
+from lsst.resources.http import _is_webdav_endpoint
```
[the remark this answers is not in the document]
airnandez: Fixed: testing of `_is_webdav_endpoint` is now included in `test_http.py`

python/lsst/resources/http.py
airnandez: You are right. I was not satisfied with that solution but thought it was the initial design and did not attempt to improve it.

I introduced a new class `SessionStore` which caches and makes reusable client sessions, one per endpoint. This gives more flexibility but there are still things to improve. For instance, the bearer token for authenticating to a remote endpoint is passed in as an environment variable `LSST_HTTP_AUTH_BEARER_TOKEN`. This means that one can only have a single token for all the remotes, which is not what happens in the practice. A similar issue happens with the client certificate to present to the server (when using X509 certificates as the client authentication method): currently one can only provide one client certificate and it is used for all the remote endpoints.

I think this should be improved in a separate ticket.

timj:
@airnandez please remember to add a change log entry in doc/changes/DM-33769.feature.rst -- a general statement of the improvements and changes to the environment variable names would suffice.

PR #8, opened 2022-02-22
DM-32250: Update setup.cfg to install py.typed file.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 jobs:
-  lint:
+  mypy:
     runs-on: ubuntu-latest
 
     steps:
--- a/python/lsst/resources/_resourcePath.py
+++ b/python/lsst/resources/_resourcePath.py
@@ -28,7 +28,21 @@ from random import Random
 
 __all__ = ("ResourcePath", "ResourcePathExpression")
 
-from typing import IO, TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Optional, Tuple, Type, Union
+from typing import (
+    IO,
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Literal,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+    overload,
+)
 
 if TYPE_CHECKING:
     from .utils import TransactionProtocol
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -29,6 +29,9 @@ tests_require =
 [options.packages.find]
 where=python
 
+[options.package_data]
+lsst.resources = py.typed
```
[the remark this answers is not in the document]
TallJimbo: Nope, `utils` had it, and that's why it worked and the only way I was able to figure out what was broken here.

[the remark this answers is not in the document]
timj: I'm even more embarrassed now then since I clearly had known at some point what I needed to do...

timj (approved):

.github/workflows/mypy.yaml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 jobs:
-  lint:
+  mypy:
```
timj, line 10: I mentioned this elsewhere but this name change means that we will have to edit the branch protections to ensure that `mypy` is an action that prevents merging. I chose `lint` deliberately so I could treat it like the other checks. I'm fine with us changing it so long as we remember to add it to the branch protections.

python/lsst/resources/_resourcePath.py
```diff
@@ -932,6 +946,18 @@ def __hash__(self) -> int:
         """Return hash of this object."""
         return hash(str(self))
 
+    def __lt__(self, other: ResourcePath) -> bool:
+        return self.geturl() < other.geturl()
```
timj, line 950: Please add some tests for these operators.

python/lsst/resources/_resourcePath.py
```diff
@@ -1103,11 +1129,41 @@ def walk(
         """
         raise NotImplementedError()
 
+    @overload
```
timj, line 1132: Thank you. I had been wondering how to handle the weirdness of this API.

python/lsst/resources/_resourcePath.py
```diff
@@ -1103,11 +1129,41 @@ def walk(
...
+    ) -> Iterator[ResourcePath]:
+        pass
+
+    @classmethod
+    def findFileResources(
+        cls,
+        candidates: Iterable[ResourcePathExpression],
+        file_filter: Optional[Union[str, re.Pattern]] = None,
         grouped: bool = False,
     ) -> Iterator[Union[ResourcePath, Iterator[ResourcePath]]]:
```
timj, line 1168: Does mypy need these type annotations if the overloads exist?

timj (approved):

python/lsst/resources/_resourcePath.py
TallJimbo: Frustratingly, it complains if I remove them entirely.  So it's either this or replacing them all with `Any`, and I think this is slightly better, in part because we've already got it.

python/lsst/resources/_resourcePath.py
```diff
@@ -1103,11 +1129,41 @@ def walk(
...
 
+    @overload
+    @classmethod
+    def findFileResources(
+        cls,
+        candidates: Iterable[ResourcePathExpression],
+        file_filter: Optional[Union[str, re.Pattern]],
+        grouped: Literal[True],
+    ) -> Iterator[Iterator[ResourcePath]]:
+        pass
```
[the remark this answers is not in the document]
TallJimbo: Oh, good point.  I had thought that was only the case for type stubs, not actual source, but I see `...` in the `typing` module reference docs examples, so I'll switch this.

PR #9, opened 2022-03-15
DM-27355: Support Google Cloud Storage

timj:
No test code at the moment.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -37,6 +37,11 @@ jobs:
         run: |
           pip install cheroot wsgidav
 
+      - name: Install google cloud storage for testing
+        shell: bash -l {0}
+        run: |
+          pip install google-cloud-storage
+
       - name: Install dependencies
         shell: bash -l {0}
         run: |
--- a/python/lsst/resources/gs.py
+++ b/python/lsst/resources/gs.py
@@ -1,0 +1,299 @@
+# This file is part of lsst-resources.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# Use of this source code is governed by a 3-clause BSD-style
+# license that can be found in the LICENSE file.
+
+"""Accessing Google Cloud Storage resources."""
+
+from __future__ import annotations
+
+__all__ = ("GSResourcePath",)
+
+import contextlib
+import logging
+import re
+import tempfile
+from typing import IO, TYPE_CHECKING, Iterator, List, Optional, Set, Tuple, Union
+
+try:
+    import google.cloud.storage as storage
+    from google.cloud.exceptions import NotFound
+except ImportError:
+    storage = None
+
+    # Must also fake the exception classes.
+    class NotFound(Exception):  # type: ignore
+        pass
+
+
+from lsst.utils.timer import time_this
+
+from ._resourcePath import ResourcePath
+
+if TYPE_CHECKING:
+    from .utils import TransactionProtocol
+
+log = logging.getLogger(__name__)
+
+
+_client = None
+"""Cached client connection."""
+
+
+def _get_client() -> storage.Client:
+    global _client
+    if storage is None:
+        raise ImportError("google-cloud-storage package not installed. Unable to communicate with GCS.")
+    if _client is None:
+        _client = storage.Client()
+    return _client
+
+
+class GSResourcePath(ResourcePath):
+    """Access Google Cloud Storage resources."""
+
+    _bucket: Optional[storage.Bucket] = None
+    _blob: Optional[storage.Blob] = None
+    _client: Optional[storage.Client] = None
+
+    @property
+    def client(self) -> storage.Client:
+        return _get_client()
+
+    @property
+    def bucket(self) -> storage.Bucket:
+        if self._bucket is None:
+            self._bucket = self.client.bucket(self.netloc)
+        return self._bucket
+
+    @property
+    def blob(self) -> storage.Blob:
+        if self._blob is None:
+            self._blob = self.bucket.blob(self.relativeToPathRoot)
+        return self._blob
+
+    def exists(self) -> bool:
+        if self.is_root:
+            return self.bucket.exists()
+        return self.blob.exists()
+
+    def size(self) -> int:
+        if self.dirLike:
+            return 0
+        # The first time this is called we need to sync from the remote.
+        # Force the blob to be recalculated.
+        try:
+            self.blob.reload()
+        except NotFound:
+            raise FileNotFoundError(f"Resource {self} does not exist")
+        size = self.blob.size
+        if size is None:
+            raise FileNotFoundError(f"Resource {self} does not exist")
+        return size
+
+    def remove(self) -> None:
+        try:
+            self.blob.delete()
+        except NotFound as e:
+            raise FileNotFoundError(f"No such resource: {self}") from e
+
+    def read(self, size: int = -1) -> bytes:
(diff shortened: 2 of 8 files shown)


python/lsst/resources/gs.py
```diff
@@ -0,0 +1,279 @@
...
+            return
+
+        # Should this method do anything at all?
+        self.blob.upload_from_string(b"")
+
+    def _as_local(self) -> Tuple[str, bool]:
+        with tempfile.NamedTemporaryFile(suffix=self.getExtension(), delete=False) as tmpFile:
+            with time_this(log, msg="Downloading %s to local file", args=(self,)):
+                try:
+                    self.blob.download_to_filename(tmpFile.name)
```
ktlim, line 130: No progress bars here, like for S3?

python/lsst/resources/gs.py
```diff
@@ -0,0 +1,279 @@
...
+        # it out locally but this could potentially lead to large memory
+        # usage for millions of keys. It will also make the initial call
+        # to this method potentially very slow. If making this method look
+        # like os.walk was not required, we could query all keys with
+        # pagination and return them in groups of 1000, but that would
+        # be a different interface since we can't guarantee we would get
+        # them all grouped properly across the 1000 limit boundary.
+        prefix = self.relativeToPathRoot if not self.is_root else ""
+        prefix_len = len(prefix)
+        dirnames: Set[str] = set()
```
ktlim, line 248: S3 has a simple list; why do these differ?

python/lsst/resources/gs.py
```diff
@@ -0,0 +1,279 @@
...
+
+        blobs = self.client.list_blobs(self.bucket, prefix=prefix, delimiter="/")
+        for page in blobs.pages:
+            # "Sub-directories" turn up as prefixes in each page.
+            dirnames.update(dir[prefix_len:] for dir in page.prefixes)
+
+            # Files are reported for this "directory" only.
+            # The prefix itself can be included as a file because we write
+            # a zero-length file for mkdir(). These must be filtered out.
+            if file_filter is not None:
```
ktlim, line 260: Seems wordier than the S3 way of doing it (get all files, then filter if necessary).

ktlim (approved):
Kind of a shame so much code needs to be repeated from `s3.py`.

python/lsst/resources/gs.py
timj: Yes, When I was copying from the S3 code I thought that it would be better to filter things out before creating the list if the list was going to be big.

python/lsst/resources/gs.py
timj: I don't suppose it's possible for duplicates so this can be a list.

python/lsst/resources/gs.py
timj: How do I do them? I didn't see a callback API for this.

python/lsst/resources/gs.py
```diff
@@ -0,0 +1,279 @@
...
+            raise ValueError(f"Bucket {self.netloc} does not exists for {self}!")
+
+        if not self.dirLike:
+            raise NotADirectoryError(f"Can not create a 'directory' for a file-like UEI {self}")
+
+        if self.is_root:
+            # The root must already exist.
+            return
+
+        # Should this method do anything at all?
```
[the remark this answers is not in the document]
timj: Yes. S3 does it and this matches that behavior. I'm not really sure what the best approach is in an object store. Make `mkdir()` a no-op and then always say a "directory" exists because in theory it could?

.github/workflows/build.yaml
```diff
@@ -37,6 +37,11 @@ jobs:
         run: |
           pip install cheroot wsgidav
 
+      - name: Install google cloud storage for testing
+        shell: bash -l {0}
+        run: |
+          pip install google-cloud-storage
```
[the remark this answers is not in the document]
timj: Yes, for the installs that are not required but are used for testing it's a toss up between doing them all at once or one at a time. If the action breaks I'll fix it.

python/lsst/resources/gs.py
```diff
@@ -0,0 +1,279 @@
...
+    def exists(self) -> bool:
+        if self.is_root:
+            return self.bucket.exists()
+        return self.blob.exists()
+
+    def size(self) -> int:
+        if self.dirLike:
+            return 0
+        # The first time this is called we need to sync from the remote.
+        # Should we track when this needs to be called again?
```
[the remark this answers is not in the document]
timj: Probably not very often for the same blob but it depends on the user. In butler we may call .size a couple of times in places.

python/lsst/resources/gs.py
```diff
@@ -0,0 +1,279 @@
...
+
+__all__ = ("GSResourcePath",)
+
+import contextlib
+import logging
+import re
+import tempfile
+from typing import IO, TYPE_CHECKING, Iterator, List, Optional, Set, Tuple, Union
+
+from google.cloud import exceptions, storage
```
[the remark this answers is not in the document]
timj: Yes. That is true for the S3 and HTTP implementations. It seemed easier than adding protection in every method but in this case it might work to put the `raise ImportError` in the `.client` property method since we always go through that. It's why pipe_base depends on boto3 (which is itself a bug because that functionality should be in ResourcePath itself).

python/lsst/resources/gs.py
ktlim: Nope, doesn't look like there is one.  You'd have to work up something with a separate thread/coroutine and `stat()`.  Not worth it.

python/lsst/resources/gs.py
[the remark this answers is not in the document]
ktlim: Since you conditionally import `gs.py`, can't you have a `try`/`except` block around this import and raise a nicer/more explanatory exception?

python/lsst/resources/gs.py
timj: Ah. It was because I was concerned that it's possible for the same prefix to turn up in multiple pages and did not want to duplicate them.

python/lsst/resources/gs.py
```diff
@@ -0,0 +1,279 @@
...
+        timer_msg = "Transfer from %s to %s"
+        timer_args = (src, self)
+
+        if isinstance(src, type(self)):
+            # Looks like a GS remote uri so we can use direct copy
+            with time_this(log, msg=timer_msg, args=timer_args):
+                rewrite_token = None
+                while True:
+                    try:
+                        rewrite_token, bytes_copied, total_bytes = self.blob.rewrite(
```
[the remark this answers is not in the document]
timj: but won't give the progress reporting that `blob.rewrite()` has?

python/lsst/resources/gs.py
```diff
@@ -0,0 +1,291 @@
...
+import logging
+import re
+import tempfile
+from typing import IO, TYPE_CHECKING, Iterator, List, Optional, Set, Tuple, Union
+
+try:
+    import google.cloud.storage as storage
+    from google.cloud.exceptions import NotFound
+except ImportError:
+    storage = None
```
ktlim, line 28: I guess due to auth considerations it isn't possible to fall back to `s3://` in this case?

timj: For the IDF the notebook setup would handle that because we have both credentials set up, although notebook environments already have google-cloud-storage installed. It would require some changes to the `__new__` method to understand that a failure to import should change the scheme to s3 but it seems like in the general case that would be more confusing than telling people to install google-cloud-storage.

python/lsst/resources/gs.py
timj: I've changed it back to the S3 way of making a big list first and then filtering it.

PR #10, opened 2022-03-22
DM-33597: Try to add 429 retry to S3

timj:
Catch the 429 client error and translate it to something
that backoff can catch.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/s3.py
+++ b/python/lsst/resources/s3.py
@@ -51,6 +51,17 @@ except ImportError:
 
     backoff = Backoff
 
+
+class _TooManyRequestsException(Exception):
+    """Private exception that can be used for 429 retry.
+
+    botocore refuses to deal with 429 error itself so issues a generic
+    ClientError.
+    """
+
+    pass
+
+
 # settings for "backoff" retry decorators. these retries are belt-and-
 # suspenders along with the retries built into Boto3, to account for
 # semantic differences in errors between S3-like providers.
@@ -64,6 +75,8 @@ retryable_io_errors = (
     # built-ins
     TimeoutError,
     ConnectionError,
+    # private
+    _TooManyRequestsException,
 )
 
 # Client error can include NoSuchKey so retry may not be the right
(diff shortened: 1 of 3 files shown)


timj:
@ktlim I've reorganized the S3 code a bit to try to make it possible to catch a 429 client error. Does this look reasonable? I haven't been able to trigger a 429 myself so I can't test it and we can't deploy ticket branch containers to the IDF.

python/lsst/resources/s3.py
```diff
@@ -114,6 +127,28 @@ def __call__(self, bytes_amount: int) -> None:
...
+    fallback : `Exception`, optional
+        Exception to raise if not intercepted. If `None` the original
+        exception will be re-raised.
+    """
+    if "(429)" in str(err):
+        # ClientError includes the error code in the message
+        # but no direct way to access it without looking inside the
+        # response.
+        raise _TooManyRequestsException(str(err)) from err
+    elif fallback:
```
ktlim, line 146: I'm not sure I like this.  Can't some ClientErrors be due to problems that are not caused by a missing object?

python/lsst/resources/s3.py
```diff
@@ -192,6 +227,22 @@ def mkdir(self) -> None:
...
+        """
+        try:
+            self.client.download_fileobj(self.netloc, self.relativeToPathRoot, local_file, Callback=progress)
+        except (
+            self.client.exceptions.NoSuchKey,
+            self.client.exceptions.NoSuchBucket,
+        ) as err:
+            raise FileNotFoundError(f"No such resource: {self}") from err
+        except ClientError as err:
+            _translate_client_error(err, FileNotFoundError(f"No such resource: {self}"))
```
ktlim, line 244: At a minimum, maybe the message should be "Resource inaccessible"?

ktlim (commented):
The restructuring seems OK, but your new fallback seems potentially problematic.

python/lsst/resources/s3.py
timj: Yes. I'm not sure how to deal with that without putting explicit handlers in the three places. I don't know the breadth of what can end up in the ClientError so I'm not sure what the possibilities are.

ktlim: Why have the fallback at all?  Just return the original ClientError if it's not translatable.

timj: The caller isn't really expecting a special S3 exception (and might not even know S3 is involved) and we mostly try to translate things. The `from err` will ensure that the previous exception is visible but given you think it's preferred I'll just put in a `raise` in the `except` block and not try to do anything special.

python/lsst/resources/s3.py
timj: I now simply re-raise.

PR #11, opened 2022-03-24
DM-34188: Use Client.bucket and not Client.get_bucket

timj:
Deriving Blobs from a bucket obtained by get_bucket seems
to require more permissions than using a blob obtained
from Client.bucket().

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/gs.py
+++ b/python/lsst/resources/gs.py
@@ -134,7 +134,7 @@ class GSResourcePath(ResourcePath):
     @property
     def bucket(self) -> storage.Bucket:
         if self._bucket is None:
-            self._bucket = self.client.get_bucket(self.netloc, retry=_RETRY_POLICY)
+            self._bucket = self.client.bucket(self.netloc)
         return self._bucket
 
     @property


python/lsst/resources/gs.py
```diff
@@ -110,7 +110,7 @@ def client(self) -> storage.Client:
     @property
     def bucket(self) -> storage.Bucket:
         if self._bucket is None:
-            self._bucket = self.client.get_bucket(self.netloc, retry=_RETRY_POLICY)
+            self._bucket = self.client.bucket(self.netloc)
```
kfindeisen, line 113: I agree that this line should not require any permissions (since [`Client.bucket` does not perform a remote operation](https://cloud.google.com/python/docs/reference/storage/latest/client#bucketbucketname-userprojectnone)). However, I don't know enough about the overall class to say whether it satisfies the requirements of DM-34188.

Would it be possible to set up a test framework that has carefully controlled permissions? You mentioned that this version works on IDF, but you also said the old version worked there.

timj: @rra if this fix looks right to you I can merge it and see if that allows you to lower your permissions request.

rra: Sounds good to me. I'm happy to test the permission change again with this change.

PR #12, opened 2022-03-29
DM-32408: Derive resources version number during build

timj:
Requires lsst/utils#119

## Checklist

- [ ] ran Jenkins


pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["setuptools", "lsst_versions @ git+https://github.com/lsst/lsst_versions@tickets/DM-32408"]
```
ktlim, line 2: Will change after merge, presumably.

ktlim (approved):

pyproject.toml
timj: Done.

PR #14, opened 2022-04-27
DM-33682: Introduce ResourceHandles

natelust:
Introduce a handle like interface to ResourcePaths. This is just a draft PR for discussion, and will be updated with new content.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
@@ -15,10 +15,10 @@ __all__ = ("BaseResourceHandle", "CloseStatus")
 from abc import ABC, abstractmethod, abstractproperty
 from io import SEEK_SET
 from logging import Logger
-from typing import Iterable, TypeVar
+from typing import Iterable, TypeVar, Union, Optional
 from enum import Enum, auto
 
-T = TypeVar('T')
+T = TypeVar('T', bound="BaseResourceHandle")
 
 
 class CloseStatus(Enum):
@@ -50,11 +50,10 @@ class BaseResourceHandle(ABC):
     _lineseperator: bytes
     _log: Logger
 
-    def __init__(self, mode, log, lineseperator=b'\n'):
+    def __init__(self, mode, log, *, newline: Union[str, bytes] = '\n'):
         self._mode = mode
         self._log = log
-        self._lineseperator = lineseperator
-        self._closed = CloseStatus.OPEN
+        self._newline = newline
 
     @property
     def mode(self) -> str:
@@ -85,15 +84,15 @@ class BaseResourceHandle(ABC):
         ...
 
     @abstractmethod
-    def readline(self, size=-1) -> bytes:
+    def readline(self, size: int = -1) -> bytes:
         ...
 
     @abstractmethod
-    def readlines(self, hint=-1) -> Iterable[bytes]:
+    def readlines(self, hint: int = -1) -> Iterable[bytes]:
         ...
 
     @abstractmethod
-    def seek(self, offset, whence=SEEK_SET) -> None:
+    def seek(self, offset: int, whence: int = SEEK_SET) -> None:
         ...
 
     @abstractmethod
@@ -105,7 +104,7 @@ class BaseResourceHandle(ABC):
         ...
 
     @abstractmethod
-    def truncate(self, size=None) -> None:
+    def truncate(self, size: Optional[int] = None) -> None:
         ...
 
     @abstractmethod
@@ -125,14 +124,15 @@ class BaseResourceHandle(ABC):
         ...
 
     @abstractmethod
-    def readinto(self, b) -> None:
+    def readinto(self, b) -> int:
         ...
 
     @abstractmethod
-    def write(self, b) -> None:
+    def write(self, b: Union[bytes, str]) -> None:
         ...
 
     def __enter__(self: T) -> T:
+        self._closed = CloseStatus.OPEN
         return self
 
     def __exit__(self, exc_type, exc_bal, exc_tb) -> None:
--- a/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
@@ -78,8 +78,8 @@ class ResourceHandleProtocol(Protocol, Generic[U]):
         ...
 
     @abstractmethod
-    def seek(self, offset: int, whence: int = SEEK_SET) -> int:
-        ...
+    def seek(self, offset: int, whence: int = SEEK_SET, /) -> int:
+        pass
 
     @abstractmethod
     def seekable(self) -> bool:
@@ -98,7 +98,7 @@ class ResourceHandleProtocol(Protocol, Generic[U]):
         ...
 
     @abstractmethod
-    def writelines(self, lines: Iterable[U]) -> None:
+    def writelines(self, lines: Iterable[U], /) -> None:
         ...
 
     @abstractmethod
@@ -106,7 +106,7 @@ class ResourceHandleProtocol(Protocol, Generic[U]):
         ...
 
     @abstractmethod
-    def write(self, b: U) -> int:
+    def write(self, b: U, /) -> int:
         ...
 
     def __enter__(self: S) -> S:
@@ -115,8 +115,9 @@ class ResourceHandleProtocol(Protocol, Generic[U]):
     def __exit__(
         self,
         exc_type: Optional[Type[BaseException]],
-        exc_bal: Optional[BaseException],
+        exc_val: Optional[BaseException],
         exc_tb: Optional[TracebackType],
+        /,
     ) -> Optional[bool]:
         ...
 
--- a/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
@@ -60,10 +60,6 @@ class BaseResourceHandle(ABC):
     def mode(self) -> str:
         return self._mode
 
-    @property
-    def name(self) -> str:
-        return ""
-
     @abstractmethod
     def close(self):
         ...
--- a/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_baseResourceHandle.py
@@ -9,16 +9,20 @@
 # Use of this source code is governed by a 3-clause BSD-style
 # license that can be found in the LICENSE file.
 from __future__ import annotations
+from types import TracebackType
 
 __all__ = ("BaseResourceHandle", "CloseStatus")
 
 from abc import ABC, abstractmethod, abstractproperty
 from io import SEEK_SET
 from logging import Logger
-from typing import Iterable, TypeVar, Union, Optional
+from typing import Callable, Iterable, TypeVar, Union, Optional, Protocol, Type, AnyStr, Generic
 from enum import Enum, auto
 
+
+S = TypeVar('S', bound="ResourceHandleProtocol")
 T = TypeVar('T', bound="BaseResourceHandle")
+U = TypeVar('U', str, bytes)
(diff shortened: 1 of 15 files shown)


python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+    def seek(self, offset: int, whence: int = SEEK_SET) -> int:
+        if self.writable():
+            if self._last_flush_position is not None:
+                if whence == SEEK_SET:
+                    offset -= self._last_flush_position
+                    if offset < 0:
+                        raise OSError("S3 ResourceHandle can not seek prior to already flushed positions")
+                if whence == SEEK_CUR:
+                    if (self.tell() - self._last_flush_position) < 0:
+                        raise OSError("S3 ResourceHandle can not seek prior to already flushed positions")
```
timj, line 200: For multipart uploading (which is what flush does) is it possible to invalidate parts that have been uploaded before the multipart upload is finalized?

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+                offset = abs(offset)
+                self._position -= offset
+        return self._position
+
+    def seekable(self) -> bool:
+        return True
+
+    def truncate(self, size: Optional[int] = None) -> int:
+        if self.writable():
+            self._buffer.truncate(size)
```
timj, line 220: What if you've flushed before you call truncate? If I call `truncate(0)` I expect to have set the file size to 0 so I can start again but it seems in this code if I call `flush(); truncate(0)` the truncate won't reset the file to 0 bytes because the buffer has been cleared by the flush and sent to S3 already. Am I missing something?

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+    def writelines(self, lines: Iterable[bytes]) -> None:
+        if self.writable():
+            self._buffer.writelines(lines)
+            self._position = self._buffer.tell()
+        else:
+            raise OSError("ResourceHandle is not writable")
+
+    def read(self, size: int = -1) -> bytes:
+        if not self.readable():
+            raise OSError("ResourceHandle is not readable")
```
timj, line 237: Sometimes you say "S3 ResourceHandle" in these errors and other times not.

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+        if "+" in self._mode:
+            self._buffer.seek(self._position)
+            return self._buffer.read(size)
+        # otherwise fetch the appropriate bytes from the remote resource
+        if size > 0:
+            stop = f"{self._position + size - 1}"
+        else:
+            stop = ""
+        args = {"Range": f"bytes={self._position}-{stop}"}
+        response = self._client.get_object(Bucket=self._bucket, Key=self._key, **args)
```
timj, line 248: See in s3.py but there are going to have to be try blocks around all the S3 API calls or backoff handlers. (there is also the `_translate_client_error` kerfuffle in that file as well.

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+        contents = response["Body"].read()
```
timj, line 249: Maybe add time_this wrapper?

python/lsst/resources/_resourcePath.py
```diff
@@ -1299,23 +1300,47 @@ def open(
...
+        encoding : `str`, optional
+            The byte encoding of any binary text
+
+        Yields
+        ------
+        handle : `BaseResourceHandle`
+            A handle that conforms to the `BaseResourcehandle interface
+        """
+        if "r" in mode or "a" in mode:
+            in_bytes = self.read()
```
timj, line 1328: The default is to read the entire file immediately? Probably should mention that in the Notes section of this implementation docstring.

timj (commented):
I have a few comments. It took me a while to get my head around it but it seems like a good start. There is plenty of missing test coverage that needs to be added (`file` URI has none) and I'll take another look when that's done as it will be easier to see how it all plays out in the context of some tests.

python/lsst/resources/_resourceHandles/_baseResourceHandle.py
```diff
@@ -0,0 +1,175 @@
...
+
+class BaseResourceHandle(ABC, ResourceHandleProtocol[U]):
+    """Base class interface for the handle like interface of `ResourcePath`
+    subclasses.
+
+    Parameters
+    ----------
+    mode : `str`
+        Handle modes as described in the python `io` module
+    log : `~logging.Logger`
```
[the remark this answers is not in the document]
natelust: I had thought this would be preferable, but I have no real opinion

python/lsst/resources/_resourceHandles/_baseResourceHandle.py
```diff
@@ -0,0 +1,175 @@
...
+        Logger to used when writing messages
+    lineseperator : `str`
```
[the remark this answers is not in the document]
natelust: Updated and resolved

python/lsst/resources/_resourceHandles/_fileResourceHandle.py
```diff
@@ -0,0 +1,107 @@
...
+    Documentation on the methods of this class line should refer to the
+    corresponding methods in the `io` module.
+    """
+
+    def __init__(
+        self, mode: str, log: Logger, *, filename: str, encoding: Optional[str], newline: str = "\n"
+    ):
+        super().__init__(mode, log, newline=newline)
+        self._filename = filename
+        # opening a file in binary mode does not support a file argument
```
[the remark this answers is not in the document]
natelust: I tried to clarify the comment and in the docstring notes

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+    possible, and all data will be buffered until the resource is closed and
+    the buffered data will be written. Additionally the entire contents of the
+    resource will be loaded into memory upon opening.
+
+    Documentation on the methods of this class line should refer to the
+    corresponding methods in the `io` module.
+    """
+
+    def __init__(
+        self, mode: str, log: Logger, client: "boto3.client", bucket: str, key: str, newline: bytes = b"\n"
```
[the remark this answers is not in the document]
natelust: I tried to clarify in the documentation

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+                # cheat a bit to get the existing data using object interfaces,
+                # because we know this is safe
+                # save the requested mode and readability
+                mode_save = self._mode
+                read_save = self._readable
+                # Update these to ensure they are strictly readable
+                self._readable = True
+                self._mode += "r"
+                self._mode = self._mode.replace("+", "")
+                self.write(self.read())
```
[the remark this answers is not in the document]
natelust: added code comment

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+            return
+        # Disallow writes to seek to a position prior to the previous flush
+        # this allows multipart uploads to upload content as the stream is
+        # written to.
+        if (
+            (self.tell() - (self._last_flush_position or 0)) < 5242880
+            and not self._closed == CloseStatus.CLOSING
+            and not self._warned
+        ):
+            warnings.warn("S3 does not support flushing objects less than 5Mb, skipping")
```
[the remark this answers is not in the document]
natelust: This warning only appears once, to let the user know that the flush operation didn't actually do anything, so they might not be surprised if nothing is uploaded, but I am happy to drop it if you feel that is better

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @property
+    def isatty(self) -> bool:
+        return False
+
+    def readable(self) -> bool:
+        return self._readable
+
+    def readline(self, size: int = -1) -> bytes:
+        raise OSError("S3 Does not support line by line reads")
```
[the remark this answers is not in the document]
natelust: The complexity of this with write, flush, and seek was large enough I didn't feel it was worth it vs the user just reading in the whole file and splitting how they chose, but if you wish I can go and have a stab at it

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
natelust: I worked through this quite a lot to find different solutions and settled on this, because after you hand off to have things uploaded it becomes a whole song and dance to find out what actually has been written out by the client, what to do if things are in flight etc. We can add that support, but I would prefer to do it on another ticket.

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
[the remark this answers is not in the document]
timj: At this point I think it's more important to get this work merged. Can you file a ticket for improving readline (and maybe add it to the message here so that people know where to comment if they hit that error).

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
timj: Ok. File a ticket.

python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
natelust: So I normally would have expected the normal course of this to be `flush(); truncate(0); flush()` if you wanted to clear out the file. That said thinking about this, it may not like that with filesize uploads. This might require more thought.

python/lsst/resources/gs.py
```diff
@@ -262,7 +264,7 @@ def open(
         *,
         encoding: Optional[str] = None,
         prefer_file_temporary: bool = False,
-    ) -> Iterator[IO]:
+    ) -> Iterator[ResourceHandleProtocol]:
```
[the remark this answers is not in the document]
natelust: hmm, my type checkers are not throwing anything, but I suppose it is because it is hidden under so many layers and ultimately import guards, this may take a bit to sort out, or I will return a Union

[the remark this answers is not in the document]
natelust: Actually, it looks like what is returned from the google code is either `TextIoWrappers` or a subclass of `io.BufferedIOBase`, which should be ok with the protocol

python/lsst/resources/_resourceHandles/_baseResourceHandle.py
```diff
@@ -0,0 +1,178 @@
...
+    corresponding methods in the `io` module.
+    """
+
+    _closed: CloseStatus
+    _mode: str
+    _log: Logger
+    _newline: U
+
+    def __init__(self, mode: str, log: Logger, *, newline: Optional[AnyStr] = None) -> None:
+        if newline is None:
```
timj, line 153: We don't have any test where `newline` is `None`?

timj:
I see that we haven't added support for WebDAV so we will need a ticket created for that since it's likely IN2P3 will want to use it.

python/lsst/resources/_resourceHandles/_baseResourceHandle.py
natelust: I dont actually think we have anything that makes use of it right now, most of the child constructors want something defined

PR #15, opened 2022-05-23
DM-34610: Disable flake8 N818 for faked exceptions.

ktlim:
This is fallout from the change to flake8 4.0 with pep8-naming 0.12.1.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/gs.py
+++ b/python/lsst/resources/gs.py
@@ -39,19 +39,19 @@ except ImportError:
     class ClientError(Exception):
         pass
 
-    class NotFound(ClientError):  # type: ignore
+    class NotFound(ClientError):  # type: ignore  # noqa: N818
         pass
 
-    class TooManyRequests(ClientError):  # type: ignore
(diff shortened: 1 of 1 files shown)


python/lsst/resources/gs.py
```diff
@@ -39,19 +39,19 @@
...
-    class NotFound(ClientError):  # type: ignore
+    class NotFound(ClientError):  # type: ignore  # noqa: N818
         pass
 
-    class TooManyRequests(ClientError):  # type: ignore
+    class TooManyRequests(ClientError):  # type: ignore  # noqa: N818
         pass
 
-    class InternalServerError(ClientError):  # type: ignore
+    class InternalServerError(ClientError):  # type: ignore  # noqa: N818
```
timj, line 48: This one doesn't need the noqa because it does end in Error.

timj (approved):
We have no choice on this one given that these are google classes we are emulating.

PR #16, opened 2022-05-26
DM-34980: Fix type annotation problems in S3

timj:
Python 3.10 mypy found additional problems.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
 
       - uses: conda-incubator/setup-miniconda@v2
         with:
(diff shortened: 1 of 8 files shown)


mwittgen:
Should I be concerned about the coverage reports failing?

mwittgen:
Should I be concerned about the coverage reports failing?

mwittgen (approved):

timj:
If you look at the coverage report you will see that it's one line that is not hit and that's a line that is a fallback line if a module is not installed. We decided we would prefer to run tests with the package installed.

PR #19, opened 2022-06-10
DM-35047: Add uploading build to PyPi

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,6 @@
-See COPYRIGHT file for code ownership.
+BSD 3-Clause License
+
+See the COPYRIGHT file for copyright information.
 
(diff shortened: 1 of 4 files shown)


LICENSE
```diff
@@ -1,4 +1,7 @@
-See COPYRIGHT file for code ownership.
+BSD 3-Clause License
+
+Copyright (c) 2022, lsst
```
timj, line 3: Where did this come from? We want people to read the COPYRIGHT file. Also `lsst` is not a legal entity.

mwittgen: lsst_versions

timj: Hilarious. That's a copy and paste error from some other package that I missed last time then.

PR #28, opened 2022-07-11
DM-35347: Add yamllint, user get_lsst_versions

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -75,7 +75,8 @@ jobs:
       - name: Run tests
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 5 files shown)


mwittgen:
yamllint crates warnings github workflows. Quoting the `on` keyword silences them.
https://github.com/adrienverge/yamllint/issues/430

.github/workflows/mypy.yaml
```diff
@@ -1,6 +1,6 @@
 name: Run mypy
 
-on:
+"on":
```
timj, line 3: We fix this in daf_butler by removing the truthiness test for github actions: https://github.com/lsst/daf_butler/blob/main/.yamllint.yaml#L14-L17

PR #33, opened 2022-10-06
DM-36412: Use atomic writes for copy or move of local files

timj:
If multiple processes are copying to the same destination we want the content of the file to be correct even if we don't know which file we end up with.

Note that os.rename does not care if the file already exists so the overwrite parameter is ignored if the target file appeared between the first check and the final rename.

## Checklist

- [ ] ran Jenkins


python/lsst/resources/file.py
```diff
@@ -284,13 +284,39 @@ def transfer_from(
...
-                ):
-                    shutil.move(local_src, newFullPath, copy_function=shutil.copy)
+                # If a rename works we try that since that is guaranteed to
+                # be atomic. If that fails we copy and rename. We do this
+                # in case other processes are trying to move to the same
+                # file and we want the "winner" to not be corrupted.
+                try:
+                    with transaction.undoWith(f"move from {local_src}", os.rename, newFullPath, local_src):
+                        os.rename(local_src, newFullPath)
+                except OSError:
```
timj, line 294: It's not easy to trigger this code path because the test environment does not have multiple disks where rename would fail.

python/lsst/resources/file.py
```diff
@@ -284,13 +284,39 @@ def transfer_from(
...
-                    f"move from {local_src}", shutil.move, newFullPath, local_src, copy_function=shutil.copy
-                ):
-                    shutil.move(local_src, newFullPath, copy_function=shutil.copy)
+                # If a rename works we try that since that is guaranteed to
+                # be atomic. If that fails we copy and rename. We do this
+                # in case other processes are trying to move to the same
+                # file and we want the "winner" to not be corrupted.
+                try:
+                    with transaction.undoWith(f"move from {local_src}", os.rename, newFullPath, local_src):
+                        os.rename(local_src, newFullPath)
```
andy-slac, line 293: If rename fails do we actually need "undo"? I'm thinking of a case, for example, when`newFullPath` already exists for some reason but you can't overwrite it, so rename fill fail, does it make sense to move existing `newFullPath` to `local_src`? Maybe if renme fails, it's safe to assume that destination did not change (or does not exist)?

python/lsst/resources/file.py
```diff
@@ -284,13 +284,39 @@ def transfer_from(
...
+                # simultaneously the file we end up with will not be corrupt.
+                with self.temporary_uri(prefix=self.parent(), suffix=self.getExtension()) as temp_copy:
+                    shutil.copy(local_src, temp_copy.ospath)
+                    with transaction.undoWith(f"copy from {local_src}", os.remove, newFullPath):
+                        # os.rename works even if the file exists.
+                        # It's possible that another process has copied a file
+                        # in whilst this one was copying. If overwrite
+                        # protection is needed then another stat() call should
+                        # happen here.
+                        os.rename(temp_copy.ospath, newFullPath)
```
andy-slac, line 319: Similar issue here, if `newFullPath` exists but cannot be overwritten, is it worth trying to delete it? You probably cannot delete it if you cannot overwrite it, but some odd conditions may still exist.

andy-slac (approved):
Looks good, though I still have questions about some corner cases, but maybe we'll never see anything that odd.

python/lsst/resources/file.py
timj: If rename fails the undo won't ever happen. `undoWith` only registers the undo callback if the code block succeeds. This is for the case where the rename works but some time later the transaction has to be rolled back.

PR #35, opened 2022-12-07
DM-37259: Accept 200 response code for PUT requests

ptrlv:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -571,7 +571,7 @@ class HttpResourcePath(ResourcePath):
 
         # Send data to its final destination.
         resp = self.put_session.put(final_url, data=data, timeout=TIMEOUT)
-        if resp.status_code not in [201, 202, 204]:
+        if resp.status_code not in [200, 201, 202, 204]:
             raise ValueError(f"Can not write file {self}, status code: {resp.status_code}")
 
 


timj:
The failure of the build is a bit odd. We haven't had a setup.py in our packages for ages. I wonder if this is a 3.8+setuptools interaction that we don't see on 3.10.

ptrlv:
Is the fix to remove the 3.8 test?

timj:
> Is the fix to remove the 3.8 test?

Try it. @mwittgen might have other thoughts.

mwittgen:
Manually restarted 3.10 build action works. Removing 3.8 and possibly 3.9 might be needed.

mwittgen:
A while ago for broken pip/setuptools versions we pinned pip<22.
The build for 3.8 in this case downgrades pip to 20, for 3.10 to version 21.
Alternatively, unpinning pip might work.

mwittgen:
I'll have a look. This is likely causing problem in other repos as well.

timj:
If we are going to drop 3.8 from the build please remember to update pyproject.toml accordingly.

mwittgen:
Fixed in: [tickets/DM-37333](https://github.com/lsst/resources/tree/tickets/DM-37333)

ptrlv:
Please check this as 3.8 is still failing. Not sure what action is needed.

timj:
@ptrlv I've rebased so we can see whether that fixes the build problem. You will have to take that into account if you pull the changes locally.

PR #38, opened 2023-02-01
DM-37510: Make HttpResourcePath.exists() more robust for WebDAV endpoints

airnandez:
## Checklist

- [x] ran Jenkins

I recognise this PR involves several changes not initially intended. The changes implemented for using webDAV-specific methods for `exist()` and `size()` had more consequences than initially anticipated. This was an opportunity (and a necessity) for improving the test suite.

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -559,7 +559,7 @@ class HttpResourcePath(ResourcePath):
                 transfer,
             )
 
-        # Short circuit if the URIs are identical immediately.
+        # Short circuit immediately if the URIs are identical.
         if self == src:
             log.debug(
                 "Target and destination URIs are identical: %s, returning immediately."
@@ -576,27 +576,19 @@ class HttpResourcePath(ResourcePath):
 
         # We can use webDAV 'COPY' or 'MOVE' if both the current and source
         # resources are located in the same server.
-        if isinstance(src, type(self)) and self.root_uri() == src.root_uri():
-            # Only available on WebDAV backends.
-            if not self.is_webdav_endpoint:
-                raise NotImplementedError(
-                    f"Endpoint {self.root_uri()} does not implement WebDAV functionality"
-                )
-
+        if isinstance(src, type(self)) and self.root_uri() == src.root_uri() and self.is_webdav_endpoint:
             with time_this(log, msg="Transfer from %s to %s directly", args=(src, self)):
-                if transfer == "move":
-                    self._move(src)
-                else:
-                    self._copy(src)
-        else:
-            # Perform the copy by downloading to a local file and uploading
-            # to the destination.
+                return self._move(src) if transfer == "move" else self._copy(src)
+
+        # For resources of different classes or for plain HTTP resources we can
+        # perform the copy or move operation by downloading to a local file
+        # and uploading to the destination.
+        with time_this(log, msg="Transfer from %s to %s via local copy", args=(src, self)):
             self._copy_via_local(src)
 
-            # This was an explicit move requested from a remote resource
-            # try to remove that resource.
-            if transfer == "move":
-                src.remove()
+        # This was an explicit move, try to remove the source.
+        if transfer == "move":
+            src.remove()
 
     def _as_local(self) -> Tuple[str, bool]:
         """Download object over HTTP and place in temporary directory.
@@ -738,6 +730,50 @@ class HttpResourcePath(ResourcePath):
         else:
             raise ValueError(f"Unable to delete resource {self}; status code: {resp.status_code}")
 
+    def _copy_via_local(self, src: ResourcePath) -> None:
+        """Replace the contents of this resource with the contents of a remote
+        resource by using a local temporary file.
+
+        Parameters
+        ----------
+        src : `HttpResourcePath`
+            The source of the contents to copy to `self`.
+        """
+        with src.as_local() as local_uri:
+            with open(local_uri.ospath, "rb") as f:
+                with time_this(log, msg="Transfer from %s to %s via local file", args=(src, self)):
+                    self._put(data=f)
+
+    def _copy_or_move(self, method: str, src: HttpResourcePath) -> None:
+        """Send a COPY or MOVE webDAV request to copy or replace the contents
+        of this resource with the contents of another resource located in the
+        same server.
+
+        Parameters
+        ----------
+        method : `str`
+            The method to perform. Valid values are "COPY" or "MOVE" (in
+            uppercase).
+
+        src : `HttpResourcePath`
+            The source of the contents to move to `self`.
+        """
+        headers = {"Destination": self.geturl()}
+        resp = self._send_webdav_request(method, url=src.geturl(), headers=headers)
+        if resp.status_code in (requests.codes.created, requests.codes.no_content):
+            return
+
+        if resp.status_code == requests.codes.multi_status:
+            tree = eTree.fromstring(resp.content)
+            status_element = tree.find("./{DAV:}response/{DAV:}status")
+            status = status_element.text if status_element is not None else "unknown"
+            error = tree.find("./{DAV:}response/{DAV:}error")
+            raise ValueError(f"{method} returned multistatus reponse with status {status} and error {error}")
+        else:
+            raise ValueError(
+                f"{method} operation from {src} to {self} failed, status code: {resp.status_code}"
+            )
+
     def _copy(self, src: HttpResourcePath) -> None:
         """Send a COPY webDAV request to replace the contents of this resource
         (if any) with the contents of another resource located in the same
@@ -760,33 +796,7 @@ class HttpResourcePath(ResourcePath):
         if must_use_local:
             return self._copy_via_local(src)
 
-        headers = {"Destination": self.geturl()}
-        resp = self._send_webdav_request("COPY", url=src.geturl(), headers=headers)
-        if resp.status_code in (requests.codes.created, requests.codes.no_content):
-            return
-
-        if resp.status_code == requests.codes.multi_status:
-            tree = eTree.fromstring(resp.content)
-            status_element = tree.find("./{DAV:}response/{DAV:}status")
-            status = status_element.text if status_element is not None else "unknown"
-            error = tree.find("./{DAV:}response/{DAV:}error")
-            raise ValueError(f"COPY returned multistatus reponse with status {status} and error {error}")
-        else:
-            raise ValueError(f"Copy operation from {src} to {self} failed, status code: {resp.status_code}")
-
-    def _copy_via_local(self, src: ResourcePath) -> None:
-        """Replace the contents of this resource with the contents of a remote
-        resource by using a local temporary file.
-
-        Parameters
-        ----------
-        src : `HttpResourcePath`
-            The source of the contents to copy to `self`.
-        """
-        with src.as_local() as local_uri:
-            with open(local_uri.ospath, "rb") as f:
-                with time_this(log, msg="Transfer from %s to %s via local file", args=(src, self)):
-                    self._put(data=f)
+        return self._copy_or_move("COPY", src)
 
     def _move(self, src: HttpResourcePath) -> None:
         """Send a MOVE webDAV request to replace the contents of this resource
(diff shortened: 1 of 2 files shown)


python/lsst/resources/http.py
```diff
@@ -373,18 +374,81 @@ def is_webdav_endpoint(self) -> bool:
...
+        <?xml version="1.0" encoding="utf-8" ?>
+        <D:propfind xmlns:D="DAV:"><D:prop><D:getlastmodified/></D:prop></D:propfind>
+        """
+        resp = self._propfind(request_body)
+        if resp.status_code == requests.codes.multi_status:  # 207
+            # Parse the XML-encoded response. Since we issue a request for a
+            # single remote resource, the response must include properties for
+            # that single resource. So either its response status is OK or not.
+            tree = etree.fromstring(resp.content.strip())
+            element = tree.find("./{DAV:}response/{DAV:}propstat/[{DAV:}status='HTTP/1.1 200 OK']")
```
ktlim, line 397: A question for my edification: can this ever be a different HTTP version (e.g. 1.0 or 2.0)?

airnandez: My understanding of the [RFC 4918](http://www.webdav.org/specs/rfc4918.html#n-propfind-status-codes)is that the response to `PROPFIND` requires that `status` be of the form `HTTP/1.1 <status code> <message>`.

But I would say this does not preclude that the communication protocol used between client and server be HTTP2 or HTTP3.

python/lsst/resources/http.py
```diff
@@ -601,6 +859,42 @@ def _openImpl(
...
+            log.debug("   body=<data>")
+        elif resp.request.body is None:
+            log.debug("   body=<empty>")
+        else:
+            log.debug(f"   body={resp.request.body[:120]!r}")
+
+        log.debug("Response:")
+        log.debug(f"   status_code={resp.status_code}")
+        log.debug(f"   headers={resp.headers}")
+        if resp.content is None:
```
timj, line 891: mypy is convinced that `resp.content` can never be `None`. I don't know if that's true but maybe changing this line to `if not resp.content` would work.

airnandez: I think mypy is wrong, indeed. I changed the test to make it happy.

timj: The problem with requests type annotations is that they come in through the `types-requests` package which is a standalone external package not directly part of the `requests` code base. It's possible there is a bug in that annotation.

airnandez: My issue is that the results of running mypy locally in my development host are not the same as the ones ran in github actions. That makes I need several roundtrips to make mypy happy. I need to take the time to understand in detail how to make it work locally.

timj: are you using the mypy from  `rubin-env-developer` conda package? I seem to have 0.991 installed and that is the version the action seems to be using.

timj:
@airnandez I rebased for you -- it was not completely trivial so it seemed unfair to make you do it. I have checked that the test_http.py file did not change after rebasing.

airnandez:
> @airnandez I rebased for you -- it was not completely trivial so it seemed unfair to make you do it. I have checked that the test_http.py file did not change after rebasing.

Thanks @timj for rebasing. I pulled the rebased branch and successfully ran the test suite. From my perspective the work I intended for this ticket is done.

python/lsst/resources/http.py
```diff
@@ -371,18 +385,85 @@ def is_webdav_endpoint(self) -> bool:
...
+            elif resp.status_code == requests.codes.not_found:
+                raise FileNotFoundError(f"Resource {self} does not exist, status code: {resp.status_code}")
+            else:
+                raise ValueError(
+                    f"Unexpected response for HEAD request for {self}, status code: {resp.status_code}"
+                )
+
+        # The remote is a webDAV server: send a PROPFIND request to retrieve
+        # the 'getcontentlength' property of the resource.
+        request_body = """
```
timj, line 442: This code is non-trivial and seems to be substantially identical to the getlastmodified implementation above. Can you consider putting this in a helper function so we only have the logic once? It seems like the code from assigning to request_body to the checks on the first `element` are the same so I think that a helper method that returned the `element` for later parsing should work okay.

python/lsst/resources/http.py
```diff
@@ -531,49 +580,269 @@ def transfer_from(
...
             transfer = self.transferDefault
 
-        if isinstance(src, type(self)):
+        # We can use webDAV 'COPY' or 'MOVE' if both the current and source
+        # resources are located in the same server.
+        if isinstance(src, type(self)) and self.root_uri() == src.root_uri():
             # Only available on WebDAV backends.
             if not self.is_webdav_endpoint:
-                raise NotImplementedError("Endpoint does not implement WebDAV functionality")
+                raise NotImplementedError(
```
timj, line 588: Is it possible to try falling back to the download/upload approach just in case PUT is allowed even if it's not WebDAV?

python/lsst/resources/http.py
```diff
@@ -531,49 +580,269 @@ def transfer_from(
...
+        redirection ourselves, since the 'requests' package changes the method
+        of the redirected request when the server responds with status 302 and
+        the method of the original request is not HEAD (which is the case for
+        webDAV requests).
+
+        That means that when the webDAV server we interact with responds with
+        a redirection to a PROPFIND or MKCOL request, the request gets
+        converted to a GET request when sent to the redirected location.
+
+        See `SessionRedirectMixin.rebuild_method()` in
```
timj, line 662: I think you need 
```
`requests.sessions.SessionRedirectMixin.rebuild_method()`
```
ie the full path. Even then it might depend on sphinx knowing where the requests documentation is located.

python/lsst/resources/http.py
```diff
@@ -531,49 +580,269 @@ def transfer_from(
...
+        request using the same HTTP verb.
+        """
+        if url is None:
+            url = self.geturl()
+
+        # Strip any whitespace at the beginning of the request body, as some
+        # XML parsers don't handle that correctly.
+        body = body.strip() if body is not None else None
+
+        for _ in range(MAX_REDIRECTS := 5):
```
timj, line 680: MAX_REDIRECTS is not a global constant so likely should not be upper case.

python/lsst/resources/http.py
```diff
@@ -531,49 +580,269 @@ def transfer_from(
...
+        # (See issues DM-37603 and DM-37651 for details)
+        #
+        # For the time being, we use a temporary local file to
+        # perform the copy client side.
+        # TODO: when those 2 issues above are solved remove the 3 lines below.
+        must_use_local = True
+        if must_use_local:
+            return self._copy_via_local(src)
+
+        headers = {"Destination": self.geturl()}
```
timj, line 769: This code from here to the end of the file is the same as the `_move` code below but with a different verb. Since this code is not used we could delete it. We could also change `def _move` to be `def _move_or_copy` and have it take an `action="MOVE"` optional parameter.

python/lsst/resources/http.py
```diff
@@ -602,6 +871,41 @@ def _openImpl(
...
+    log.debug("Request")
+    log.debug(f"   method={resp.request.method}")
+    log.debug(f"   URL={resp.request.url}")
+    log.debug(f"   headers={resp.request.headers}")
+    if resp.request.method == "PUT":
+        log.debug("   body=<data>")
+    elif resp.request.body is None:
+        log.debug("   body=<empty>")
+    else:
+        log.debug(f"   body={resp.request.body[:120]!r}")
```
timj, line 896: Do not use f-strings in `log.debug` calls. This should be:

```python
log.debug("   body=%r", resp.request.body[:120])
```

tests/test_http.py
```diff
@@ -47,160 +47,225 @@ def setUp(self):
...
-        self.notExistingFileResourcePath = ResourcePath(
-            f"https://{serverRoot}/{existingFolderName}/{notExistingFileName}"
+
+        # DAV existing folder and its parent directory
+        self.davExistingFolderResource = ResourcePath(
+            f"{self.davEndpoint}/{existingFolderName}", forceDirectory=True
         )
-        self.existingFolderResourcePath = ResourcePath(
-            f"https://{serverRoot}/{existingFolderName}", forceDirectory=True
+        href = self.davExistingFolderResource.geturl().replace(
```
timj, line 69: Isn't this code the same as doing `href = self.davExistingFolderResource.relativeToPathRoot` ?

tests/test_http.py
```diff
@@ -47,160 +47,225 @@ def setUp(self):
...
+            content_type="text/xml; charset=utf-8",
+            auto_calculate_content_length=True,
         )
-        self.handleWithRangeResourcePath = ResourcePath(
-            f"https://{serverRoot}/{existingFolderName}/{fileForHandleWithRange}"
+
+        href = (
+            self.davExistingFolderResource.parent()
+            .geturl()
+            .replace(str(self.davExistingFolderResource.root_uri()), "", 1)
```
timj, line 101: Again this is `.relativeToPathRoot`.

tests/test_http.py
```diff
@@ -47,160 +47,225 @@ def setUp(self):
...
-            responses.HEAD,
-            self.handleWithRangeResourcePath.geturl(),
-            status=200,
-            headers={"Content-Length": "1024", "Accept-Ranges": "true"},
+        # DAV existing file.
+        self.davExistingFileResource = ResourcePath(
+            f"{self.davEndpoint}/{existingFolderName}/{existingFileName}"
+        )
+        href = self.davExistingFileResource.geturl().replace(
+            str(self.davExistingFileResource.root_uri()), "", 1
```
timj, line 134: I will not comment on this again.

timj (changes requested):
Thanks for doing all this work and significantly enhancing the test code.

I have some minor cleanups I think are needed.

python/lsst/resources/http.py
```diff
@@ -74,7 +75,8 @@ def _is_webdav_endpoint(path: Union[ResourcePath, str]) -> bool:
     _is_webdav_endpoint : `bool`
         True if the endpoint implements WebDAV, False if it doesn't.
     """
-    if (ca_cert_bundle := os.getenv("LSST_HTTP_CACERT_BUNDLE")) is None:
+    is_https = str(path).lower().startswith("https://")
```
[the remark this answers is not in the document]
airnandez: Fixed. This check is actually not needed here.

python/lsst/resources/http.py
```diff
@@ -74,7 +75,8 @@ def _is_webdav_endpoint(path: Union[ResourcePath, str]) -> bool:
...
+    if (ca_cert_bundle := os.getenv("LSST_HTTP_CACERT_BUNDLE")) is None and is_https:
```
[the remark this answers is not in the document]
airnandez: Fixed. The intention was to give some hints to the user in case of authentication errors. Hints are provided when actual authentication errors are found.

python/lsst/resources/http.py
```diff
@@ -84,7 +86,18 @@ def _is_webdav_endpoint(path: Union[ResourcePath, str]) -> bool:
     log.debug("Detecting HTTP endpoint type for '%s'...", path)
     verify: Union[bool, str] = ca_cert_bundle if ca_cert_bundle else True
     resp = requests.options(str(path), verify=verify)
-    return "DAV" in resp.headers
+
+    # Explicitly check that "1" is part of the value of the "DAV" header
+    # since some servers liberally put information in that header than
```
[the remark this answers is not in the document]
airnandez: Fixed.

python/lsst/resources/http.py
```diff
@@ -84,7 +86,18 @@ def _is_webdav_endpoint(path: Union[ResourcePath, str]) -> bool:
...
+    # is not required by the RFC.
+    if "DAV" not in resp.headers:
+        return False
+    else:
+        dav_version = resp.headers.get("DAV")
```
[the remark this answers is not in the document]
airnandez: This was over defensive. Improved.

python/lsst/resources/http.py
```diff
@@ -84,7 +86,18 @@ def _is_webdav_endpoint(path: Union[ResourcePath, str]) -> bool:
...
+        if dav_version is None:
+            return False
+        else:
+            return "1" in dav_version.replace(" ", "").split(",")
```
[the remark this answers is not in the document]
airnandez: Fixed: example provided and code simplified.

python/lsst/resources/http.py
airnandez: Refactored. Introduced helper class and helper function to parse `PROPFIND` responses.

python/lsst/resources/http.py
airnandez: Fixed. `transfer_from` using plain HTTP servers now use a download-then-upload operation.

python/lsst/resources/http.py
airnandez: Fixed.

python/lsst/resources/http.py
airnandez: Fixed. Introduced helper method `copy_or_move()`.

python/lsst/resources/http.py
airnandez: Fixed.

tests/test_http.py
airnandez: Yes, the result is almost the same, although not always. Fixed.

tests/test_http.py
airnandez: Fixed (same as above).

tests/test_http.py
airnandez: Fixed (same as above).

timj (approved):
Thanks for making the changes. Looks great.

airnandez:
This was far much work than I initially anticipated. But I am happy with the result. I think this code is more robust and overall better now. Thanks for your time and for this very detailed feedback.

PR #41, opened 2023-02-22
DM-37523: Implement walk() for HttpResourcePath class

airnandez:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -1149,7 +1149,7 @@ class DavProperty:
         for propstat in response.findall("./{DAV:}propstat"):
             # Only extract properties of interest with status OK.
             status = propstat.find("./{DAV:}status")
-            if status is None or not self._status_ok_rex.match(status.text):
+            if status is None or not self._status_ok_rex.match(str(status.text)):
                 continue
 
             for prop in propstat.findall("./{DAV:}prop"):
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -1170,27 +1170,27 @@ class DavProperty:
                     self._displayname = str(element.text).strip()
 
     @property
-    def exists(self):
+    def exists(self) -> bool:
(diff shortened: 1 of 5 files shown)


python/lsst/resources/http.py
```diff
@@ -682,19 +742,35 @@ def _send_webdav_request(
...
-                return resp
+        with time_this(
+            log,
+            msg="%s %s",
+            args=(
+                method,
+                url,
+            ),
+            mem_usage=True,
+            mem_unit=u.mebibyte,
```
timj, line 753: Are you worried about memory usage?

python/lsst/resources/http.py
```diff
@@ -27,6 +27,7 @@
 from typing import TYPE_CHECKING, BinaryIO, Iterator, List, Optional, Tuple, Union, cast
 
 import requests
+from astropy import units as u
```
timj, line 30: This requires that we add `astropy` to the `https` optional dependencies section in `pyproject.toml`.

tests/test_http.py
```diff
@@ -227,10 +228,26 @@ def test_dav_as_local(self):
         self.assertEqual(ResourcePath(local_path).read(), contents)
         os.remove(local_path)
 
+    def test_dav_size(self):
+        # Size of an inexistent file must raise
```
timj, line 232: "a nonexistent file"  (also, period at end of comment)

timj (approved):
Looks great. Thanks. Minor comments.

python/lsst/resources/http.py
airnandez: The intention is that we collect as much resource consumption information as we can when we are timing the execution. I thought it would not cost too much since `time_this` does not collect the memory usage when the logger is not enable for  `DEBUG` level, which is the normal production configuration.

tests/test_http.py
airnandez: Fixed.

PR #44, opened 2023-03-03
DM-37439: Fix FileReadWriteTestCase test

Fireye04:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/s3utils.py
+++ b/python/lsst/resources/s3utils.py
@@ -11,7 +11,13 @@
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from unittest import TestCase
+
 __all__ = (
+    "clean_test_environment",
     "getS3Client",
     "s3CheckFileExists",
     "bucketExists",
@@ -106,6 +112,27 @@ retryable_client_errors = (
 # are not included.
 all_retryable_errors = retryable_io_errors
 max_retry_time = 60
+
+
+def clean_test_environment(testcase: TestCase) -> None:
+    """Clear S3_ENDPOINT_URL then restore it at the end of a test.
+
+    Parameters
+    ----------
+    testcase: `unittest.TestCase`
+        Reference to the test being run; used to add a cleanup function.
+    """
+    endpoint = os.environ.get("S3_ENDPOINT_URL")
+
+    if not endpoint:
+        return
+    os.environ["S3_ENDPOINT_URL"] = ""
+
+    def cleanup() -> None:
+        if endpoint is not None:
+            os.environ["S3_ENDPOINT_URL"] = endpoint
+
+    testcase.addCleanup(cleanup)
 
 
 def getS3Client() -> boto3.client:
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -9,6 +9,7 @@
 # Use of this source code is governed by a 3-clause BSD-style
 # license that can be found in the LICENSE file.
 
+import os
 import unittest
 
 from lsst.resources import ResourcePath
@@ -43,6 +44,12 @@ class S3ReadWriteTestCase(GenericReadWriteTestCase, unittest.TestCase):
         # Enable S3 mocking of tests.
         self.mock_s3.start()
 
+        self.endpoint = None
+
+        if "S3_ENDPOINT_URL" in os.environ:
+            self.endpoint = os.environ["S3_ENDPOINT_URL"]
+            os.environ["S3_ENDPOINT_URL"] = ""
+
         # set up some fake credentials if they do not exist
         # self.usingDummyCredentials = setAwsEnvCredentials()
 
@@ -66,6 +73,9 @@ class S3ReadWriteTestCase(GenericReadWriteTestCase, unittest.TestCase):
 
         bucket = s3.Bucket(self.netloc)
         bucket.delete()
+
+        if self.endpoint is not None:
+            os.environ["S3_ENDPOINT_URL"] = self.endpoint
 
         # Stop the S3 mock.
         self.mock_s3.stop()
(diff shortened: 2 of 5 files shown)


tests/test_s3.py
```diff
@@ -43,6 +44,16 @@ def setUp(self):
         # Enable S3 mocking of tests.
         self.mock_s3.start()
 
+        self.endpoint = None
+
+        if "S3_ENDPOINT_URL" in os.environ:
+            self.endpoint = os.environ["S3_ENDPOINT_URL"]
+            os.environ["S3_ENDPOINT_URL"] = ""
+
+        for k, v in os.environ.items():
```
timj, line 53: You need to remove these debugging lines

Fireye04: whoops! I'll get on that

python/lsst/resources/tests.py
```diff
@@ -28,6 +28,19 @@
 TESTDIR = os.path.abspath(os.path.dirname(__file__))
 
 
+def prepare_testing(testcase: unittest.TestCase) -> None:
```
timj, line 31: Since this is specifically for use by S3 tests only I would be inclined to put this in `s3utils` and not the generic `tests.py`. Pondering the name of the function, I think I would prefer something like `clean_test_environment`. You will need to add to `__all__` when you move it.

tests/test_s3.py
```diff
@@ -43,6 +43,14 @@ def setUp(self):
         # Enable S3 mocking of tests.
         self.mock_s3.start()
 
+        prepare_testing(self)
+
+        # self.endpoint = None
```
timj, line 48: These commented lines have to go away.

tests/test_s3.py
```diff
@@ -67,6 +75,9 @@ def tearDown(self):
         bucket = s3.Bucket(self.netloc)
         bucket.delete()
 
+        # if self.endpoint is not None:
```
timj, line 78: Remove.

python/lsst/resources/tests.py
```diff
@@ -28,6 +28,19 @@
...
+    endpoint = os.environ.get("S3_ENDPOINT_URL")
+
+    if not endpoint:
+        return
+    os.environ["S3_ENDPOINT_URL"] = ""
+
+    def cleanup() -> None:
+        os.environ["S3_ENDPOINT_URL"] = Optional[endpoint]
```
timj, line 39: This looks wrong. `Optional[]` is a type annotation and is not something that should be assigned.

Mypy is complaining that the variable `endpoint` is `str | None` and so telling you that you can't assign `None` to `os.environ` which is defined as `dict[str, str]`. You can fix this by changing the code to be:

```python
if endpoint:
    os.environ["S3_ENDPOINT_URL"] = endpoint
```

timj (changes requested):

python/lsst/resources/tests.py
timj: That should say `if endpoint is not None` or course.

python/lsst/resources/s3utils.py
```diff
@@ -11,7 +11,10 @@
 
 from __future__ import annotations
 
+from unittest import TestCase
```
timj, line 14: The normal way of dealing with the need to import some code that is only used for type checking is to do something like:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from unittest import TestCase
```

python/lsst/resources/s3utils.py
```diff
@@ -108,6 +111,20 @@ class _TooManyRequestsException(Exception):
 max_retry_time = 60
 
 
+def clean_test_environment(testcase: TestCase) -> None:
```
timj, line 114: There will need to be a docstring here that explains the purpose of this function.

python/lsst/resources/tests.py
```diff
@@ -10,7 +10,10 @@
 # license that can be found in the LICENSE file.
 from __future__ import annotations
 
-__all__ = ["GenericTestCase", "GenericReadWriteTestCase"]
+__all__ = [
```
timj, line 13: This reformatting seems odd. I think what happened is you added a new function to the end of the list, that triggered black to reformat it, then you removed the new function but black sees no reason to change it back to how it was. To reduce git history churn can you please take it back to how it was and make sure to squash the commit history so the change never happens?

tests/test_s3.py
```diff
@@ -43,6 +44,8 @@ def setUp(self):
         # Enable S3 mocking of tests.
         self.mock_s3.start()
 
+        clean_test_environment(self)
```
timj, line 47: I'm ambivalent as to whether this line needs a comment before it explaining that we do this to be safe against S3_ENDPOINT_URL is set. The function name seems clear enough so as long as we have a docstring in that function that explains its purpose I think it's fine.

timj (approved):
Looks good. Needs some minor fixes.

timj (changes requested):
I hit the wrong button last time 😄

python/lsst/resources/s3utils.py
```diff
@@ -108,6 +114,27 @@ class _TooManyRequestsException(Exception):
...
+    """Clear S3_ENDPOINT_URL then reset it at the end of a test.
```
timj, line 118:
```suggestion
    """Clear S3_ENDPOINT_URL then restore it at the end of a test.
```

python/lsst/resources/s3utils.py
```diff
@@ -108,6 +114,27 @@ class _TooManyRequestsException(Exception):
...
+
+    Parameters
+    ----------
+    testcase: `TestCase`, `self`
```
timj, line 122:
```suggestion
    testcase: `unittest.TestCase`
```

timj (approved):
Looks okay. I've made a couple of suggestions to the docstring.

PR #45, opened 2023-03-09
DM-38156: Improve persistent connection handling for HttpResourcePath class

airnandez:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -69,7 +69,7 @@ class HttpReadWriteWebdavTestCase(GenericReadWriteTestCase, unittest.TestCase):
         # keep network connections to the remote server open and have no
         # means through the API exposed by Requests of actually close the
         # underlyng sockets to make tests pass without warning.
-        warnings.filterwarnings(action="ignore", message="unclosed", category=ResourceWarning)
+        warnings.filterwarnings(action="ignore", message=r"unclosed.*socket", category=ResourceWarning)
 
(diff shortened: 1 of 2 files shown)


tests/test_http.py
```diff
@@ -476,6 +496,110 @@ def _is_server_running(cls, port: int) -> bool:
                 return False
 
 
+class HttpResourcePathConfigTestCase(unittest.TestCase):
+    """Test for the HttpResourcePathConfig class."""
+
+    def test_send_expect_header(self):
+        # Ensure environment variable LSST_HTTP_PUT_SEND_EXPECT_HEADER is
+        # inspected to initialize the HttpResourcePath config class.
+        os.environ.pop("LSST_HTTP_PUT_SEND_EXPECT_HEADER", None)
```
timj, line 505: In theory a test should not change the environment permanently in a way that might affect other tests.  It looks like below you use `unittest.mock.patch.dict()` with `clear=True`. Couldn't you also use that approach here?

```python
with unittest.mock.patch.dict(os.environ, {}, clear=True):
    config = HttpResourcePathConfig()
    self.assertFalse(config.send_expect_on_put)
```
?

timj (approved):
This looks great. I didn't read every line and I'm not an expert on WebDAV but it looked good to me. Some minor comments. Unfortunately the tests don't run any faster.

PR #47, opened 2023-04-04
DM-38552: Support non-file root in constructor for relative paths

timj:
Also fix the strange behavior where absolute schemeless URIs can sometimes be updated to `file` URIs but not always.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourcePath.py
+++ b/python/lsst/resources/_resourcePath.py
@@ -85,11 +85,14 @@ class ResourcePath:
     root : `str` or `ResourcePath`, optional
         When fixing up a relative path in a ``file`` scheme or if scheme-less,
         use this as the root. Must be absolute.  If `None` the current
-        working directory will be used. Can be a file URI.
+        working directory will be used. Can be a file URI. Not used if
+        ``forceAbsolute`` is `False`.
     forceAbsolute : `bool`, optional
         If `True`, scheme-less relative URI will be converted to an absolute
         path using a ``file`` scheme. If `False` scheme-less URI will remain
-        scheme-less and will not be updated to ``file`` or absolute path.
+        scheme-less and will not be updated to ``file`` or absolute path unless
+        it is already an absolute path, in which case it will be updated to
+        a ``file`` scheme.
     forceDirectory: `bool`, optional
         If `True` forces the URI to end with a separator, otherwise given URI
         is interpreted as is.
--- a/python/lsst/resources/_resourcePath.py
+++ b/python/lsst/resources/_resourcePath.py
@@ -748,7 +748,9 @@ class ResourcePath:
             return None
         if self.netloc != other.netloc:
             # Special case for localhost vs empty string.
-            if {self.netloc, other.netloc} != {"", "localhost"}:
+            # There can be many variants of localhost.
+            local_netlocs = {"", "localhost", "localhost.localdomain", "127.0.0.1"}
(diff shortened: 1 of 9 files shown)


python/lsst/resources/_resourcePath.py
```diff
@@ -85,11 +85,14 @@ class ResourcePath:
     root : `str` or `ResourcePath`, optional
```
andy-slac, line 85: `uri` description above says "Can be scheme-less if referring to a local filesystem path", does it need to be updated to say scheme-less is relative to `root` (which does not have to be a local path)?

python/lsst/resources/_resourcePath.py
```diff
@@ -85,11 +85,14 @@ class ResourcePath:
...
         When fixing up a relative path in a ``file`` scheme or if scheme-less,
         use this as the root. Must be absolute.  If `None` the current
-        working directory will be used. Can be a file URI.
+        working directory will be used. Can be a file URI. Not used if
+        ``forceAbsolute`` is `False`.
```
andy-slac, line 89: What happens if `uri` is a relative with `file` scheme (`file:a/b.c`) but `root` uses a different scheme? Does it make sense to support something like this, or will it fail with an exception?

python/lsst/resources/_resourcePath.py
```diff
@@ -259,7 +297,7 @@ def __new__(
                 )
 
             parsed, dirLike = subclass._fixupPathUri(
-                parsed, root=root, forceAbsolute=forceAbsolute, forceDirectory=forceDirectory
+                parsed, root=root_uri, forceAbsolute=forceAbsolute, forceDirectory=forceDirectory
```
andy-slac, line 300: If `parsed` is already joined with `root_uri`, should `root_uri` be set to `None` before sending it to this method?

python/lsst/resources/_resourcePath.py
```diff
@@ -697,8 +735,12 @@ def relative_to(self, other: ResourcePath) -> Optional[str]:
             other = other.abspath()
 
         # Scheme-less self is handled elsewhere.
-        if self.scheme != other.scheme or self.netloc != other.netloc:
+        if self.scheme != other.scheme:
             return None
+        if self.netloc != other.netloc:
+            # Special case for localhost vs empty string.
+            if {self.netloc, other.netloc} != {"", "localhost"}:
```
andy-slac, line 742: What about more esoteric options, e.g. `localhost.localdomain`? 🙂

andy-slac (approved):
Looks good, few minor questions.

python/lsst/resources/_resourcePath.py
timj: `file:a/b.c` is not really even allowed -- it's not a valid URI but `urllib.parse` does support it. I currently convert it to an absolute URI because a `file` URI must be an absolute path and now that scheme-less implies "relative path" and can be relative to anything I either have to convert this to an absolute file URI or raise an exception because I don't have a "relative schemeless that MUST BE A FILE" any more.

python/lsst/resources/_resourcePath.py
timj: This is working because `root` parameter will be ignored because the path is now an absolute path and also the subclass is not a file URI scheme any more. I agree it would be clearer to set it to None once we've used it above.

PR #49, opened 2023-04-05
DM-38589: Fix repeated reads with stream handle

timj:
This works around a bug found in wsgidav where reading byte ranges past the end of file return the entire file contents and not 416 status code (see mar10/wsgidav#281).

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
@@ -152,5 +152,5 @@ class HttpReadResourceHandle(BaseResourceHandle[bytes]):
             self._completeBuffer.seek(0)
             return self.read(size=size)
 
-        self._current_position += size
+        self._current_position += len(resp.content)
         return resp.content
--- a/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
@@ -140,7 +140,8 @@ class HttpReadResourceHandle(BaseResourceHandle[bytes]):
 
         if (code := resp.status_code) not in (200, 206):
             raise FileNotFoundError(
-                f"Unable to read resource {self._url}, or bytes are out of range; status code: {code}"
+                f"Unable to read resource {self._url}, or byte request {self._current_position}-{end_pos}"
+                f" is out of range; status code: {code}"
             )
 
         # verify this is not actually the whole file and the server did not lie
--- a/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
@@ -181,14 +181,17 @@ class HttpReadResourceHandle(BaseResourceHandle[bytes]):
         # server.
         if "Content-Range" in resp.headers:
             content_range = resp.headers["Content-Range"]
-            _, range_string = content_range.split(" ")
-            range, total = range_string.split("/")
-            if "-" in range:
-                _, end = range.split("-")
-                end_pos = int(end)
-                if total != "*":
-                    if end_pos == int(total) - 1:
-                        self._eof = True
+            units, range_string = content_range.split(" ")
+            if units == "bytes":
+                range, total = range_string.split("/")
+                if "-" in range:
+                    _, end = range.split("-")
+                    end_pos = int(end)
+                    if total != "*":
+                        if end_pos == int(total) - 1:
+                            self._eof = True
+            else:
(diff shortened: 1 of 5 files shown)


python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -122,34 +131,70 @@ def read(self, size: int = -1) -> bytes:
...
+        # or a read has previously been done. Any time we specify a byte range
+        # we must disable the gzip compression on the server since we want
+        # to address ranges in the uncompressed file. If we send ranges that
+        # are interpreted by the server as offsets into the compressed file
+        # then that is at least confusing and also there is no guarantee that
+        # the bytes can be uncompressed.
 
         end_pos = self._current_position + (size - 1) if size >= 0 else ""
-        headers = {"Range": f"bytes={self._current_position}-{end_pos}"}
+        headers = {"Range": f"bytes={self._current_position}-{end_pos}", "Accept-Encoding": ""}
```
rra, line 150: I think technically setting this to `identity` is more correct (I'm not sure exactly what requests does with empty headers under the hood).

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -122,34 +131,70 @@ def read(self, size: int = -1) -> bytes:
...
 
-        with time_this(self._log, msg="Read from remote resource %s", args=(self._url,)):
+        with time_this(
+            self._log, msg="Read from remote resource %s using headers %s", args=(self._url, headers)
+        ):
             resp = self._session.get(self._url, stream=False, timeout=self._timeout, headers=headers)
 
-        if (code := resp.status_code) not in (200, 206):
+        if resp.status_code == requests.codes.range_not_satisfiable:
```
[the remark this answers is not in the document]
rra: This behavior is required by the standard. See https://httpwg.org/specs/rfc9110.html#byte.ranges.

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -122,34 +131,70 @@ def read(self, size: int = -1) -> bytes:
...
+        # server.
+        if "Content-Range" in resp.headers:
+            content_range = resp.headers["Content-Range"]
+            _, range_string = content_range.split(" ")
+            range, total = range_string.split("/")
+            if "-" in range:
+                _, end = range.split("-")
+                end_pos = int(end)
+                if total != "*":
+                    if end_pos == int(total) - 1:
```
rra, line 190: I would use `>=` here just out of paranoia. (I can imagine some situation where the file was truncated while the web server was sending it, although I admit it's incredibly unlikely.)

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -63,7 +65,9 @@ def fileno(self) -> int:
         raise io.UnsupportedOperation("HttpReadResourceHandle does not have a file number")
 
     def flush(self) -> None:
-        raise io.UnsupportedOperation("HttpReadResourceHandles are read only")
+        modes = set(self._mode.split())
+        if {"w", "x", "a", "+"} & modes:
+            raise io.UnsupportedOperation("HttpReadResourceHandles are read only")
```
rra, line 70: This doesn't look right, although maybe `self._mode` isn't a normal Python mode string? `split()` splits on whitespace, so a mode string of `w+` or `wb` would test as false in this test, but it looks like the intent is for it to be true.

rra:
Not sure why GitHub added some of my comments more than once....

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
timj: I love it when tests pass without a problem despite the code being completely wrong. I see that `modes = set(self._mode)` does work just fine...

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
timj: Thanks. @airnandez mentioned exactly this on Slack this morning so I will change it.

timj:
> Not sure why GitHub added some of my comments more than once....

@rra I don't see doubles but there are two comments you made that I got in email but which are no longer visible to me on the PR so maybe you thought there were doubles and deleted some but they weren't really doubles?

rra (approved):
Looks good! I am surprised that Python doesn't allow `.split("")`; that's kind of annoying from a clarity standpoint, but this certainly works.

PR #51, opened 2023-04-08
DM-38599: Fix contents length checking in server-encoded HTTP responses

airnandez:
## Checklist

- [x] ran Jenkins


timj (approved):
Looks okay. I don't think the two length variable definitions needed to move but I'm also fine with them in the new location.

PR #52, opened 2023-04-10
DM-38589: (Alternative) Fix EOF for HTTP file handle

timj:
This is an alternative implementation of #49:

- It uses the Content-Range header to find the end (which works around the wsgidav bug (mar10/wsgidav#281).
- It falls back to comparing the size read to the requested size to indicate EOF.
- It disables Accept-Encoding when using byte ranges (since otherwise the byte ranges are for the compressed file).

The first change in particular allows the repeated use of read() to not need additional calls to the server off the end of the byte range.

I will integrate this back into #49 once reviewed.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
@@ -152,5 +152,5 @@ class HttpReadResourceHandle(BaseResourceHandle[bytes]):
             self._completeBuffer.seek(0)
             return self.read(size=size)
 
-        self._current_position += size
+        self._current_position += len(resp.content)
         return resp.content
--- a/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
@@ -63,7 +63,11 @@ class HttpReadResourceHandle(BaseResourceHandle[bytes]):
         raise io.UnsupportedOperation("HttpReadResourceHandle does not have a file number")
 
     def flush(self) -> None:
-        raise io.UnsupportedOperation("HttpReadResourceHandles are read only")
+        modes = set(self._mode.split())
+        if {"w", "x", "a", "+"} & modes:
+            raise io.UnsupportedOperation("HttpReadResourceHandles are read only")
+        if self._completeBuffer is not None:
+            self._completeBuffer.flush()
 
     @property
     def isatty(self) -> Union[bool, Callable[[], bool]]:
--- a/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
@@ -140,7 +140,8 @@ class HttpReadResourceHandle(BaseResourceHandle[bytes]):
 
         if (code := resp.status_code) not in (200, 206):
             raise FileNotFoundError(
-                f"Unable to read resource {self._url}, or bytes are out of range; status code: {code}"
+                f"Unable to read resource {self._url}, or byte request {self._current_position}-{end_pos}"
+                f" is out of range; status code: {code}"
             )
 
         # verify this is not actually the whole file and the server did not lie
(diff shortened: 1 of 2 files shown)


python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -63,7 +65,11 @@ def fileno(self) -> int:
         raise io.UnsupportedOperation("HttpReadResourceHandle does not have a file number")
 
     def flush(self) -> None:
-        raise io.UnsupportedOperation("HttpReadResourceHandles are read only")
+        modes = set(self._mode.split())
+        if {"w", "x", "a", "+"} & modes:
+            raise io.UnsupportedOperation("HttpReadResourceHandles are read only")
+        if self._completeBuffer is not None:
+            self._completeBuffer.flush()
```
natelust, line 72: This seems ok, but really it is just an implementation detail, flush on a read only handle is not supposed to do anything. I had set it to an unsupported operation because I didnt want people to think it did anything, but I think we mention in enough places handles are read only that it is fine.

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -129,27 +140,64 @@ def read(self, size: int = -1) -> bytes:
...
-        with time_this(self._log, msg="Read from remote resource %s", args=(self._url,)):
+        with time_this(
+            self._log, msg="Read from remote resource %s using headers %s", args=(self._url, headers)
+        ):
             resp = self._session.get(self._url, stream=False, timeout=self._timeout, headers=headers)
 
+        if resp.status_code == 416:
+            # Must have run off the end of the file. A standard file handle
+            # will treat this as EOF so be consistent with that.
+            self._current_position = int(end_pos)
```
natelust, line 162: I dont think int(end_pos) here is correct, that is saying it is at the position past the end of the file, as if it could read that far. Is it true that every 416 will have what bytes it can read returned? I thought some would send back a 416 with no bytes read sometimes? If that is true, then you do need to ask again for the remaining bytes.

natelust (changes requested):
Request more for comment and or justification.

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
timj: I was trying to comply with what happens when you try to do this with a normal file handle. When  you call seek() off the end of the file and then call read() you end up with empty bytes as if EOF and tell() reports your seek position. Now that I write this I realize that not changing the current position at all would be consistent so I will remove this line.

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
timj: We have to make it be allowed. Forwarding on the flush seemed like the best way of being consistent but I can leave it as a no-op if you like.

natelust: I think it is fine, it just that the _completeBuffer only is used in some code paths (when a full read is done) and I wanted to highlight that only in those paths is forwarding it on meaningful, perhaps a simple comment would be in order

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
natelust: I am also worried about the case where:
1) the current position is at byte 10 of a 12 byte file
2) someone request a read of 4 bytes
3) a 416 is returned, but those last 2 bytes may not have been read.

natelust: Ok, it seemed a little ambiguous when I read https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416

timj: That does seem to be ambiguous but github and wsgidav are definitely letting you ask for 4 bytes but only returning 2.

timj:
Work on this PR has been moved to #49.

PR #53, opened 2023-04-11
DM-38492: Move ResourcePathExpression definition to end of file

timj:
This allows it to be a real definition rather than relying on a string to specify the ResourcePath entry. Doing it this away allows users to use ResourcePathExpression in an isinstance call.

## Checklist

- [ ] ran Jenkins


andy-slac (approved):
Looks good.

PR #54, opened 2023-04-17
DM-38742: Enable use of Ceph multi-tenant bucket names.

ktlim:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/s3utils.py
+++ b/python/lsst/resources/s3utils.py
@@ -33,11 +33,13 @@ __all__ = (
 
 import functools
 import os
+import re
 from http.client import HTTPException, ImproperConnectionState
 from types import ModuleType
 from typing import Any, Callable, Optional, Tuple, Union, cast
 
 from botocore.exceptions import ClientError
+from botocore.handlers import validate_bucket_name
 from urllib3.exceptions import HTTPError, RequestError
 
 try:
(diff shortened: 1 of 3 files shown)


python/lsst/resources/s3utils.py
```diff
@@ -147,6 +148,10 @@ def getS3Client() -> boto3.client:
     -----
     The endpoint URL is from the environment variable S3_ENDPOINT_URL.
     If none is specified, the default AWS one is used.
+
+    If the environment variable LSST_CEPH_BUCKETS exists (with any value),
```
timj, line 152: I'm concerned that CEPH is not the only reason that someone would want to do this. How about calling it something like `LSST_S3_DISABLE_VALIDATION` ? Also, I'm not sure I like it that we don't have to set it to a truthy value.

tests/test_s3utils.py
```diff
@@ -82,6 +85,16 @@ def testBucketExists(self):
         self.assertTrue(bucketExists(f"{self.bucketName}"))
         self.assertFalse(bucketExists(f"{self.bucketName}_no_exist"))
 
+    def testCephBucket(self):
```
timj, line 88: Does this test fail if someone has set the environment variable in the shell?

timj (changes requested):
Looks okay in theory but I'd like a different name and I worry about env var existence vs truth/false value.

timj (approved):
Looks okay. Thanks.

PR #56, opened 2023-04-25
DM-38858: Avoid resetting the sessions unnecessary to prevent leaking socket de…

airnandez:
…scriptors

## Checklist

- [x] ran Jenkins


timj (approved):
This looks okay. I need to think a bit more how this leaked in a situation where there was no forked subprocess.

airnandez:
I added more details in JIRA to keep a record of this bug.

PR #57, opened 2023-05-04
DM-39044: Add as_local support for Python resource URIs

timj:
## Checklist

- [ ] ran Jenkins


python/lsst/resources/packageresource.py
```diff
@@ -39,6 +42,31 @@ class PackageResourcePath(ResourcePath):
...
+            except ModuleNotFoundError:
+                pass
+            else:
+                dirLike = ref.is_dir()
+        return parsed, dirLike
+
+    def isdir(self) -> bool:
+        """Return True if this URI is a directory, else False."""
+        ref = resources.files(self.netloc).joinpath(self.relativeToPathRoot)
+        # The item might not exist yet.
```
TallJimbo, line 67: The comment in the method above this one suggests that it'll raise `ModuleNotFoundError`, right?  I guess I'm not sure what this comment is saying.

python/lsst/resources/packageresource.py
```diff
@@ -21,13 +23,14 @@
 else:
     import importlib_resources as resources  # type: ignore[no-redef]
 
-from typing import Iterator
-
-__all__ = ("PackageResourcePath",)
+from typing import TYPE_CHECKING, Iterator, Tuple
```
TallJimbo, line 26:
```suggestion
from collections.abc import Iterable
from typing import TYPE_CHECKING
```

TallJimbo (approved):

python/lsst/resources/packageresource.py
timj: Yes, when I wrote this code I had no idea that `ModuleNotFoundError` was a possibility...

python/lsst/resources/packageresource.py
```diff
@@ -39,6 +42,31 @@ class PackageResourcePath(ResourcePath):
     resource name.
     """
 
+    @classmethod
+    def _fixDirectorySep(
+        cls, parsed: urllib.parse.ParseResult, forceDirectory: bool = False
+    ) -> Tuple[urllib.parse.ParseResult, bool]:
```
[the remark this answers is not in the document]
timj: I continually convince myself that that doesn't work on python 3.8 and then I continually remember that we only run mypy on python 3.10.

python/lsst/resources/packageresource.py
timj: I've refactored to do that check in one place and added a test.

PR #59, opened 2023-06-29
DM-39832: Clean up sphinx docs

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/lsst.resources/index.rst
+++ b/doc/lsst.resources/index.rst
@@ -31,3 +31,9 @@ Python API reference
 
 .. automodapi:: lsst.resources
    :no-main-docstr:
+
+.. automodapi:: lsst.resources.utils
+   :no-main-docstr:
+
+.. automodapi:: lsst.resources._resourceHandles._baseResourceHandle
+   :no-main-docstr:
--- a/doc/lsst.resources/index.rst
+++ b/doc/lsst.resources/index.rst
@@ -32,8 +32,13 @@ Python API reference
 .. automodapi:: lsst.resources
    :no-main-docstr:
 
-.. automodapi:: lsst.resources.utils
-   :no-main-docstr:
 
-.. automodapi:: lsst.resources._resourceHandles._baseResourceHandle
-   :no-main-docstr:
+.. _lsst.resources-internal-api:
+
+Internal API reference
+======================
+
+.. toctree::
+   :maxdepth: 1
+
+   internal-api.rst
--- a/doc/lsst.resources/index.rst
+++ b/doc/lsst.resources/index.rst
@@ -33,12 +33,9 @@ Python API reference
    :no-main-docstr:
 
 
-.. _lsst.resources-internal-api:
-
-Internal API reference
-======================
+See also :ref:`lsst.resources-internal-api`.
 
 .. toctree::
-   :maxdepth: 1
+   :hidden:
 
    internal-api.rst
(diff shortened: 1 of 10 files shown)


doc/lsst.resources/index.rst
```diff
@@ -31,3 +31,9 @@ Python API reference
 
 .. automodapi:: lsst.resources
    :no-main-docstr:
+
+.. automodapi:: lsst.resources.utils
+   :no-main-docstr:
+
+.. automodapi:: lsst.resources._resourceHandles._baseResourceHandle
+   :no-main-docstr:
```
TallJimbo, line 39: I'm surprised this directive doesn't put those private namespaces into the docs.  Or does it actually do that, and it's still the least bad way to get the docs for the classes in at all?

python/lsst/resources/_resourceHandles/_baseResourceHandle.py
```diff
@@ -38,10 +38,10 @@ class CloseStatus(Enum):
 
 class ResourceHandleProtocol(Protocol, Generic[U]):
     """Defines the interface protocol that is compatible with children of
-    `BaseResourceHandle`.
+    `.BaseResourceHandle`.
```
TallJimbo, line 41: <rant>
Ugh, I see so much of this in our future.  I really wish Sphinx could just look at how we've imported symbols!
</rant>

python/lsst/resources/_resourcePath.py
```diff
@@ -1299,7 +1299,7 @@ def open(
 
         Returns
         -------
-        cm : `contextlib.ContextManager`
+        cm : `contextlib.contextmanager`
```
TallJimbo, line 1302: This is a function, not a type.  I think you want `contextlib.AbstractContextManager`

TallJimbo (approved):

doc/lsst.resources/index.rst
timj: They do appear in the docs in their own subsection. The problem is that if you specify private base classes then none of the links from your classes resolve any where unless sphinx has built the docs for the private base class. Maybe there is a directive here that would let the class docs exist but not list them on the main page.

python/lsst/resources/_resourcePath.py
timj: or maybe change it to Yields an iterator of type resource handle protocol.

doc/lsst.resources/index.rst
TallJimbo: Would propagating just the relevant class up to the package namespace via `__all__` work?  Not sure it'd actually be better.

timj: Yes. These aren't exported by default so they aren't indexed. The automodapi forces them to be rendered so that they can be cross-referenced. I don't think we really want to have a huge public interface in the python layer to make sphinx's job easier. Maybe we can create a subpage for "private interfaces" and then automodapi them there.

python/lsst/resources/_resourcePath.py
TallJimbo: I don't think so; the use of a decorator that changes the return type should be an implementation detail as far as the docs are returned.  MyPy is smart enough to transform the return type via the decorator but Sphinx cannot.

doc/lsst.resources/index.rst
TallJimbo: I like the "private interfaces" (or maybe "internal interfaces"?) subpage idea.

timj: Done that. It's not quite how I'd want it (since it currently shows a list of internal modules on the main page) but better than it was.

python/lsst/resources/_resourcePath.py
timj: Ok. Will change it back. In much of our other documentation I think we say what the context manager yields and don't declare the decorated thing.

PR #63, opened 2023-08-17
DM-40247: Improve webdav interface as suggested by ruff S checks

airnandez:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -422,7 +422,7 @@ class BearerTokenAuth(AuthBase):
 
     def __call__(self, req: requests.PreparedRequest) -> requests.PreparedRequest:
         # Only add a bearer token to a request when using secure HTTP.
-        if req.url.lower().startswith("https://") and self._token:
+        if req.url and req.url.lower().startswith("https://") and self._token:
             self._refresh()
             req.headers["Authorization"] = f"Bearer {self._token}"
         return req
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -307,15 +307,16 @@ def _is_webdav_endpoint(path: ResourcePath | str) -> bool:
     try:
         session = requests.Session()
         session.mount(str(path), HTTPAdapter(max_retries=retries))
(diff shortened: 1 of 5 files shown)


timj:
To fix the mypy problem add something like:

```ini
[mypy-defusedxml.*]
ignore_missing_imports = True
```

This PR works with both xml and defusedxml so we do not have to wait on the rubin-env RFC.

Also, there is no reason why we can't add defusedxml as a requirement in the `pyproject.toml` and `requirements.txt` now for the pypi release and GitHub actions.

airnandez:
Thanks @timj for your feedback. I added `defusedxml` to the appropriate files and now all the tests pass.

python/lsst/resources/http.py
```diff
@@ -235,38 +265,88 @@ def _is_webdav_endpoint(path: ResourcePath | str) -> bool:
...
-                f"Unexpected response to OPTIONS request for {path}, status: {resp.status_code} {resp.reason}"
+        session = requests.Session()
+        session.mount(str(path), HTTPAdapter(max_retries=retries))
+        session.verify = os.environ.get("LSST_HTTP_CACERT_BUNDLE", True)
+        with session:
+            resp = session.options(
+                str(path),
+                stream=False,
+                timeout=(
+                    _timeout_from_environment("LSST_HTTP_TIMEOUT_CONNECT", 30.0),
```
timj, line 316: Should we be reading the environment variables here if we are saying that OPTIONS should use a shorter timeout than other activities?

timj (approved):
Looks great. Thanks.

python/lsst/resources/http.py
airnandez: I thought using the timeout values specified by the user via the environment variables for `OPTIONS` as well would be better for consistency. In addition, using the environment allows the user to configure the environment for using a potential slow server, as opposed to hardcoding a magic value.

timj: Yes, okay. I was concerned that people might want different values for OPTIONS and other things but if the env var is only used for testing it doesn't matter.

PR #75, opened 2024-01-12
tickets/PREOPS-4765: Add option to lsst.resources.S3ResourcePath to disable threading

ehneilsen:
## Checklist

- [X] ran Jenkins
- [X] added a release note for user-visible changes to `doc/changes`

### DIFF
--- a/python/lsst/resources/s3.py
+++ b/python/lsst/resources/s3.py
@@ -22,7 +22,7 @@ import sys
 import tempfile
 import threading
 from collections.abc import Iterable, Iterator
-from functools import cached_property
+from functools import cache, cached_property
 from typing import IO, TYPE_CHECKING, cast
 
 from botocore.exceptions import ClientError
@@ -122,6 +122,32 @@ def _translate_client_error(err: ClientError) -> None:
         raise FileNotFoundError("Resource not found: {self}")
 
 
+@cache
+def _parse_string_to_maybe_bool(maybe_bool_str: str) -> bool | None:
+    """Map a string to either a boolean value or None.
+
+    Parameters
+    ----------
+    maybe_bool_str : `str`
+        The value to parse
+
+    Results
+    -------
+    maybe_bool : `bool` or `None`
+        The parsed value.
+    """
+    if maybe_bool_str.lower() in ["t", "true", "yes", "y", "1"]:
+        maybe_bool = True
+    elif maybe_bool_str.lower() in ["f", "false", "no", "n", "0"]:
+        maybe_bool = False
+    elif maybe_bool_str.lower() in ["none", ""]:
+        maybe_bool = None
+    else:
+        raise ValueError(f'Value of "{maybe_bool_str}" is not True, False, or None.')
+
+    return maybe_bool
+
+
 class S3ResourcePath(ResourcePath):
     """S3 URI resource path implementation class.
 
@@ -145,18 +171,9 @@ class S3ResourcePath(ResourcePath):
         try:
             use_threads_str = os.environ["LSST_S3_USE_THREADS"]
         except KeyError:
-            use_threads_str = None
+            use_threads_str = "None"
 
-        if use_threads_str is None:
-            use_threads = None
-        elif use_threads_str.lower() in ["t", "true", "yes", "y", "1"]:
-            use_threads = True
-        elif use_threads_str.lower() in ["f", "false", "no", "n", "0"]:
-            use_threads = False
-        elif use_threads_str.lower() in ["none", ""]:
-            use_threads = None
-        else:
-            raise ValueError(f'LSST_S3_USE_THREADS value of "{use_threads_str}" is not True, False, or None.')
+        use_threads = _parse_string_to_maybe_bool(use_threads_str)
 
         return use_threads
 
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -13,6 +13,7 @@ import os
 import time
 import unittest
 from inspect import signature
+from unittest import mock
 from urllib.parse import parse_qs, urlparse
 
 from lsst.resources import ResourcePath
@@ -77,13 +78,6 @@ class S3ReadWriteTestCase(GenericReadWriteTestCase, unittest.TestCase):
 
         # Stop the S3 mock.
         self.mock_s3.stop()
-
-        # Make user environ is cleaned up after testing different values
-        # of LSST_S3_USE_THREADS
-        try:
-            del os.environ["LSST_S3_USE_THREADS"]
-        except KeyError:
-            pass
 
         S3ResourcePath.use_threads = None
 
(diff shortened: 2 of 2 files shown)


python/lsst/resources/s3.py
```diff
@@ -118,6 +125,25 @@ def _translate_client_error(err: ClientError) -> None:
 class S3ResourcePath(ResourcePath):
     """S3 URI resource path implementation class."""
 
+    use_threads: bool | None = None
```
timj, line 128: Can you add a doc string here please? (triple quoted string underneath).

python/lsst/resources/s3.py
```diff
@@ -118,6 +125,25 @@ def _translate_client_error(err: ClientError) -> None:
...
+
+    @property
+    def _transfer_config(self) -> TransferConfig:
+        if self.use_threads is None:
+            try:
+                # Use yaml.safe_load to be flexible with what is accepted.
+                self.use_threads = bool(yaml.safe_load(os.environ["LSST_S3_USE_THREADS"]))
```
timj, line 135: I'm a bit torn on using yaml here to parse a string when we are in control of the environment variable setting and can document it to be 0 or 1. Do we really have to support "no" and "false" as well?

python/lsst/resources/s3.py
```diff
@@ -118,6 +125,25 @@ def _translate_client_error(err: ClientError) -> None:
 class S3ResourcePath(ResourcePath):
     """S3 URI resource path implementation class."""
```
timj, line 126: Please document the environment variable here in a notes section like we do in `HttpResourcePath`.

python/lsst/resources/s3.py
```diff
@@ -118,6 +125,25 @@ def _translate_client_error(err: ClientError) -> None:
...
+
+    @property
+    def _transfer_config(self) -> TransferConfig:
+        if self.use_threads is None:
+            try:
+                # Use yaml.safe_load to be flexible with what is accepted.
+                self.use_threads = bool(yaml.safe_load(os.environ["LSST_S3_USE_THREADS"]))
+            except KeyError:
+                # Set it to None to use boto's default.
+                self.use_threads = None
```
timj, line 138: Isn't this setting the instance value and not the class property?

tests/test_s3.py
```diff
@@ -164,6 +165,11 @@ def _check_presigned_url(self, url: str, expiration_time_seconds: int):
         # out to lunch for a while on a busy CI machine
         self.assertLessEqual(abs(expected_expiration_timestamp - actual_expiration_timestamp), 120)
 
+    def test_unthreaded_local(self):
+        S3ResourcePath.use_threads = False
+        self.test_local()
```
timj, line 170: If this test fails then `use_threads` will be stuck in `False`.

The test is tricky because you can't really test that anything changed inside boto, especially with moto mocking out much of boto.

timj (commented):
Before I sign off on this, I am wondering about the class vs instance property. Did you mean for this behavior to be modifiable per instance? The news fragment and the test code implies not, but it looks like as written you can change it per instance. This is tricky with `ResourcePath` in general since the purpose is that people don't really know where the resource is pointing to and so a per-instance property really needs to work in the base class for all resources.

ehneilsen:
I did think about the class vs. instance property thing. As it is now, when setting `use_threads`, the user get to choose whether they want to affect just the instance or all instances of the class for which they have not set it explicitly by choosing whether to assign `my_instance.use_threads` or `S3ResourceClass.use_threads`, just as they can make that choice for any other attribute (because that's all it is).
For example, if I had two instances of `S3ResourcePath`, `a` and `b`, and I assigned `a.use_threads` to some value, I would find it pretty surprising if it affected `b` at all, because that's not how attributes in python normally work. On the other hand, if I assigned `S3ResourcePath.use_threads`, I would not be surprised if it affected both.
If I assign `S3ResourcePath.use_threads` and it did not affect `a` because I had previously assigned `a.use_threads` and forgotten about it, I might find it a little surprising, but only a little, because this is a pretty normal python gatcha, occurring for any attribute.

tests/test_s3.py
ehneilsen: I'll go ahead and reset use_threads in the tearDown, which should fix that.

python/lsst/resources/s3.py
```diff
@@ -116,7 +122,50 @@ def _translate_client_error(err: ClientError) -> None:
...
+    within threads other than python's main thread.
+    """
+
+    use_threads: bool | None = None
+    """Explicitly turn on or off threading in use of boto's download_fileobj.
+    Setting this to None results in boto's default behavior."""
+
+    @property
+    def _transfer_config(self) -> TransferConfig:
+        if self.use_threads is None:
```
timj, line 144: Doesn't this block get executed every single time by default? Even if the environment variable parsing hasn't changed and the environment variable itself is the same? How about:

```python
_read_use_threads_environ = False
_cached_use_threads = None
def _read_use_threads_from_environment() -> bool:
    if _read_use_threads_environ:
        return _cached_use_threads
    try:
        s3_use_threads = os.environ[...]
    except KeyError:
        s3_use_threads = None
    if s3_use_threads is None:
        ...
    elif ...:
        ...

    _read_use_threads_environ = True
    _cached_use_threads = use_threads
    return use_threads
```
and then call this from `_transfer_config` ? It seems like the environment parsing is no longer a quick thing because you want to support a myriad of different values rather than 0 1 or unset.

ehneilsen: How about:
```

    @cached_property
    def _environ_use_threads(self):
        try:
            use_threads_str = os.environ["LSST_S3_USE_THREADS"]
        except KeyError:
            use_threads_str = None

        if use_threads_str is None:
            use_threads = None
        elif use_threads_str.lower() in ["t", "true", "yes", "y", "1"]:
            use_threads = True
        elif use_threads_str.lower() in ["f", "false", "no", "n", "0"]:
            use_threads = False
        elif use_threads_str.lower() in ["none", ""]:
            use_threads = None
        else:
            raise ValueError(
                f'LSST_S3_USE_THREADS value of "{use_threads_str}" is not True, False, or None.'
            )

        return use_threads

    @property
    def _transfer_config(self) -> TransferConfig:
        if self.use_threads is None:
            self.use_threads = self._environ_use_threads

        if self.use_threads is None:
            transfer_config = TransferConfig()
        else:
            transfer_config = TransferConfig(use_threads=self.use_threads)

        return transfer_config
```

timj: Works for me. I keep forgetting about cached_property.

tests/test_s3.py
```diff
@@ -164,6 +176,33 @@ def _check_presigned_url(self, url: str, expiration_time_seconds: int):
...
+
+    def test_explicit_default_threading(self):
+        os.environ["LSST_S3_USE_THREADS"] = "None"
+        S3ResourcePath.use_threads = None
+        boto_default = signature(boto3.s3.transfer.TransferConfig).parameters["use_threads"].default
+        test_resource_path = self.root_uri.join("test_file.dat")
+        self.assertEqual(test_resource_path._transfer_config.use_threads, boto_default)
+
+    def test_threading_false(self):
+        os.environ["LSST_S3_USE_THREADS"] = "False"
```
timj, line 199: The tests make it clear that we aren't caching the parsing of the environment variable because otherwise this test would fail because it would be cached from a previous read. `@cached_property` is per-instance, so we are reparsing every time. I'm a bit sensitive to this at the moment because I'm looking at overheads for transferring 800k files and things that don't take long add up when there are that many ResourcePath objects.

PS The normal thing in tests is to use:

```python
with mock.patch.dict(os.environ, {"LSST_S3_USE_THREADS": "False"}):
    ...
```

ehneilsen: Or I can pull the parsing code into a new function (outside of the class), say _parse_use_threading, use a plain cached decorator on it.  The code would then still check the environ var for each instance (but still not each use of each instance), but the parsing code would only happen once (per value the environ takes during the execution, probably only one except for the tests).

timj: Ok. Try that.

timj (approved):
Thanks for the changes.

I still worry about that `resource.use_threads` only works if S3 is used but we can worry about that in the future. I don't really want to end up with a lot of user code that checks `resource.scheme`.

PR #76, opened 2024-01-16
DM-42306: Performance optimizations

timj:
* No longer call os.path.isdir when constructing a schemeless URI.
* Simplify relativeToPathRoot and getExtension to remove Path constructor.
* Add tri-state for dirLike and forceDirectory to allow someone to declare that a URI is definitely file-like or else say that they don't know.

## Checklist

- [ ] ran Jenkins


python/lsst/resources/file.py
```diff
@@ -123,7 +123,10 @@ def isdir(self) -> bool:
             `True` if this URI is a directory or looks like a directory,
             else `False`.
         """
-        return self.dirLike or os.path.isdir(self.ospath)
+        if self.dirLike is None:
+            # Cache state for next time.
+            self.dirLike = os.path.isdir(self.ospath)
```
timj, line 128: `ResourcePath` is nominally immutable but this cache can be updated from `None` to `bool`.  This will presumably need locking around it and in packageresource and schemeless implementations.

I am also wondering if this and isTemporary should be made read-only getters with `_dirLike` used internally.

python/lsst/resources/_resourcePath.py
```diff
@@ -602,15 +616,22 @@ def getExtension(self) -> str:
             indicating file compression, in which case the combined
             extension (e.g. ``.fits.gz``) will be returned.
         """
+        if self.isdir():
+            return ""
+
```
dhirving, line 621: For schemeless and file forceDirectory=None, this now calls `stat` where previously no I/O was necessary.

python/lsst/resources/_resourcePath.py
```diff
@@ -1360,7 +1402,7 @@ def open(
         required to reimplement `open`, though they may delegate to `super`
         when ``prefer_file_temporary`` is `False`.
         """
-        if self.dirLike:
+        if self.isdir():
             raise IsADirectoryError(f"Directory-like URI {self} cannot be opened.")
```
dhirving, line 1406: You might consider removing this check entirely -- won't the underlying open() implementation throw an appropriate error if we try to open a directory as a file?

tests/test_schemeless.py
```diff
@@ -70,6 +70,26 @@ def test_creation(self) -> None:
         with self.assertRaises(ValueError):
             ResourcePath(relative, root=ResourcePath("resource://lsst.resources/something.txt"))
 
+    def test_isdir(self):
```
dhirving, line 73: Unless I'm way off base on my comments above, I think we may be missing some test coverage for the other ResourcePath methods in the schemeless forceDirectory=None case.

python/lsst/resources/_resourcePath.py
```diff
@@ -485,22 +493,24 @@ def parent(self) -> ResourcePath:
         For a file-like URI this will be the same as calling `dirname()`.
         """
         # When self is file-like, return self.dirname()
-        if not self.dirLike:
+        if not self.isdir():
```
dhirving, line 496: forceDirectory=None adds a call to stat where previously no I/O was needed

dhirving (changes requested):
I like the general thrust of this change, but I'm concerned about some of the details.

Unless I'm misunderstanding things, this seems to break some of the code paths for schemeless paths when the user does not explicitly specify forceDirectory, in the cases where schemeless is used to represent "some path that will be later bound to a root other than the current working directory".

It also appears to add many new calls to `stat` for File and Schemeless paths where previously no I/O was done, if the user doesn't explicitly specify forceDirectory.

python/lsst/resources/schemeless.py
```diff
@@ -46,9 +47,10 @@ def isabs(self) -> bool:
         Returns
         -------
         isabs : `bool`
-            `True` if the file is absolute, `False` otherwise.
+            `True` if the file is absolute, `False` otherwise. Will always
+            be `False` for schemeless URIs.
         """
-        return os.path.isabs(self.ospath)
+        return False
```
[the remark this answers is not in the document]
timj: No. In the past it was possible for a schemeless URI to be absolute, but now they are always upgraded to `file` schemes and a schemeless URI can always return `False` here.

```
>>> from lsst.resources import ResourcePath
>>> ResourcePath("/a/b/c.txt")
ResourcePath("file:///a/b/c.txt")
>>> ResourcePath("/a/b/c.txt", forceAbsolute=False)
ResourcePath("file:///a/b/c.txt")
```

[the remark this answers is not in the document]
timj: It's possible that there is some vestigial code in this file that didn't get cleaned up when the change was made in #47.

python/lsst/resources/schemeless.py
```diff
@@ -70,9 +72,38 @@ def abspath(self) -> ResourcePath:
...
+        if self.dirLike is None:
+            try:
+                status = os.stat(self.ospath)
+            except FileNotFoundError:
+                # Do not update dirLike flag.
+                return False
+
+            # Cache state for next time.
+            self.dirLike = stat.S_ISDIR(status.st_mode)
+        return self.dirLike
```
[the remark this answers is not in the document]
timj: This is no different to how `os.path.isdir` works in that if you give it a relative path it will look for it relative to cwd.  At issue I think is whether `dirLike is None` should ever cache the result since it becomes invalidated if you change directory.

The dirLike annotation lets you avoid relying on the cwd which, as you say, can get a bit confusing if `dir1` is a file now but if I move to a different directory it's a directory.  

Butler now does force the issue when it knows there is a file but there are some cases where the user gives us a schemeless URI (for say the location of butler root) and the thing using it doesn't know whether it's a directory or a file. Most days I will do `butler query-datasets ../pipelines_check/DATA` and butler has to work out that I meant it to use the `butler.yaml` inside `DATA` and didn't mean that I wanted it to load a file called `DATA`. The `isdir` check here with `dirLike is None` triggers in this case (and zsh has a habit of removing the trailing `/` on tab complete and so the `ResourcePath` constructor can't work it out from the trailing `/`).

python/lsst/resources/_resourcePath.py
```diff
@@ -206,7 +211,7 @@ def __new__(
             # We invoke __new__ again with str(self) to add a scheme for
             # forceAbsolute, but for the others that seems more likely to paper
             # over logic errors than do something useful, so we just raise.
-            if forceDirectory and not uri.dirLike:
+            if forceDirectory and not uri.isdir():
```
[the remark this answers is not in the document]
timj: If `forceDirectory and uri.dirLike is False` should definitely raise an exception.

The question is what you do if the URI you are given doesn't actually know it's a directory or not. If you are doing

```
resource_copy = ResourcePath(resource, forceDirectory=True)
```
and `resource` doesn't know if it's a directory or not, then things can get dicey if it is in fact a local file. But yes, if this happens to be a relative path that will soon be attached to an S3 root URI then checking the file system for directory-ness is not great. Maybe we only do this check if it's a `file` URI since then we really can tell and now the `os.stat` call will only happen if it hasn't already been cached or pre-declared. Then for schemeless we cross fingers and hope you know what you are doing.

We do have a problem in that we can't tell the difference between schemeless that we know is going to be attached to a root at some point and schemeless that is a local file relative path.

python/lsst/resources/_resourcePath.py
```diff
@@ -239,7 +244,7 @@ def __new__(
                     and root_uri.scheme != "file"  # file scheme has different code path
                     and not parsed.path.startswith("/")  # Not already absolute path
                 ):
-                    if not root_uri.dirLike:
+                    if not root_uri.isdir():
                         raise ValueError(
                             f"Root URI ({root}) was not a directory so can not be joined with"
                             f" path {parsed.path!r}"
```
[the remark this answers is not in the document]
timj: root_uri is required to be a directory. This code is going to trigger a stat call in rare case when root is a schemeless URI itself. All other cases dirLike will only be a boolean. Again it becomes an issue of "assume it's a directory because what other choice is there" vs "is that directory a real directory relative to cwd".

python/lsst/resources/_resourcePath.py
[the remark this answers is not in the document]
dhirving: I really think that instantiating a ResourcePath should never involve doing any I/O, both for performance and because many use cases involve building up paths that will exist in the future but don't exist yet.

What exactly gets dicey?  Most of the path manipulation code I've seen is able to work without accessing the filesystem, by making assumptions about what is and is not a directory. e.g. if you are trying to join "a/b/c" to "d/e/f", you just assume that 'c' is a directory and return "a/b/c/d/e/f" because why else would you be trying to access files inside it?  Then in the unlikely event that someone created a file named 'c' where there was supposed to be a directory, you get an exception when you try to access it as a directory.

The really tricky cases are things like abspath() that have to know what is and isn't a symlink, but we're already deferring to the python library for those from what I've seen.

python/lsst/resources/_resourcePath.py
```diff
@@ -426,6 +430,10 @@ def split(self) -> tuple[ResourcePath, str]:
         Equivalent to `os.path.split` where head preserves the URI
         components.
         """
+        if self.isdir():
+            # This is a directory so must return itself and the empty string.
+            return self, ""
+
```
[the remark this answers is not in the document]
timj: For `a/b` it will return `a/b` for the directory if both `a` and `b` happen to exist relative to cwd and `b` is a directory. This does mean that you can get a different answer based on what directory you happen to be sitting in for schemeless. Changing semantics such that people are required to provide the trailing slash for a directory is going to break things a little as described above with zsh dropping `/` on autocomplete on the command line.

os.path and pathlib assume things based on a trailing slash that they sometimes drop in other scenarios.

python/lsst/resources/_resourcePath.py
timj: No. Prevoiusly it always required the stat on construction, now it requires the stat the first time it needs to know.

python/lsst/resources/_resourcePath.py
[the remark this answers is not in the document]
timj: I believe we have resolved this during standup: try not to be clever given we don't really know if the schemeless URI has any relationship to the local file system at all.

python/lsst/resources/_resourcePath.py
```diff
@@ -418,24 +429,29 @@ def split(self) -> tuple[ResourcePath, str]:
...
-            unquoted.
+            separator or if the URI is known to be associated with a directory.
+            Tail will never contain separators. It will be unquoted.
 
         Notes
         -----
         Equivalent to `os.path.split` where head preserves the URI
         components.
         """
+        if self.dirLike:
```
timj, line 440: One problem remaining is this (which you can reproduce from the resources directory):

```python
>>> from lsst.resources import ResourcePath
>>> r = ResourcePath("python")
>>> r
ResourcePath("file:///Users/timj/work/lsstsw/build/resources/python")
>>> r.parent()
ResourcePath("file:///Users/timj/work/lsstsw/build/resources/")
>>> r.dirname()
ResourcePath("file:///Users/timj/work/lsstsw/build/resources/")
>>> r.isdir()
True
>>> r.dirname()
ResourcePath("file:///Users/timj/work/lsstsw/build/resources/python")
>>> r.parent()
ResourcePath("file:///Users/timj/work/lsstsw/build/resources/")
```
so `dirname()` gives the right answer for the absolute file URI only when it's allowed to check the file system to be sure and the answer changes depending on whether `isdir` has been called or not.

dhirving: You might consider removing dirname() entirely.  I just looked and outside of Butler the only thing using it is lsst.alert.packet.schema:from_uri, which only uses it for remote URLs and would get identical results from parent().  I haven't looked but I strongly suspect the Butler cases (other than the ButlerConfig thing) would be just as happy using parent() as well.

dhirving: From my point of view dirname is just "parent, but confusing."

timj: `os.path.dirname` is clearly ridiculous given it changes what it does based on whether it thinks it's a directory or not. `Path.parent` doesn't care.

python/lsst/resources/_resourcePath.py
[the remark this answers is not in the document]
timj: Offline we have agreed to leave `split()` calling `isdir()` and to deprecate `split`, `dirname`, and `basename` on another ticket. `pathlib.Path` has a consistent approach of `parent` and `name` and we will switch to that and if someone wants to know if it is a directory they call `isdir` themselves.

python/lsst/resources/_resourcePath.py
timj: This one made no sense at all so it now returns the extension regardless of whether it's a directory or not.

python/lsst/resources/_resourcePath.py
```diff
@@ -541,6 +551,10 @@ def updatedFile(self, newfile: str) -> ResourcePath:
         Forces the ResourcePath.dirLike attribute to be false. The new file
         path will be quoted if necessary.
         """
+        if self.isdir():
+            # This is a directory so we can append the new file directly.
+            return self.join(newfile)
+
```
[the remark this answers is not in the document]
timj: I changed it to dirLike and added a comment that we need to deprecate this current behavior and make it always use parent (which is what `PosixPath.with_name()` does).

python/lsst/resources/_resourcePath.py
timj: Hopefully as_local is not a hotspot. 

* For `file` URIs this is a no-op so it will let a directory through and you'll get a deferred error sometime later when you call read().
* For S3 you will currently get a zero-byte file which will successfully be downloaded and which you can open.

So at minimum the dirLike test is needed (at least until S3 stops writing 0-byte directories), but given this is about to potentially involve downloading lots of data the isdir() check seems fairly harmless.

tests/test_schemeless.py
timj: The full coverage report indicates that we are in the high 90s% for schemeless and resource path. I can't see any obvious gaps from the report.

python/lsst/resources/schemeless.py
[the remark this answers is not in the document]
timj: I removed the cache setting so that cwd will do the right thing if you are explicitly asking if a relative path refers to a local directory.

python/lsst/resources/_resourcePath.py
timj: Changed it back to look at `dirLike` as an optimization rather than something that changes behavior (because the profile showed that PosixPath has a long construction time).

timj:
@dhirving  I think this is ready for review again. I fixed the butler issue (it was related to the updatedFile changes). I think the main outcomes are:

* Need to deprecated `dirname`, `split`, and `basename`
* Need to deprecate `updatedFile` deciding to do a `join` if it sees a directory.

dhirving (approved):
Looks good to me

PR #77, opened 2024-01-17
DM-42522: Implement exists() and size() for S3 presigned URLs

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
+++ b/python/lsst/resources/_resourceHandles/_httpResourceHandle.py
@@ -14,9 +14,10 @@ from __future__ import annotations
 __all__ = ("HttpReadResourceHandle",)
 
 import io
+import re
 from collections.abc import Callable, Iterable
 from logging import Logger
-from typing import AnyStr
+from typing import AnyStr, NamedTuple
 
 import requests
 from lsst.utils.timer import time_this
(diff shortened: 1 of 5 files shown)


dhirving:
There are a few lines with missing coverage which were already missing coverage in the existing code.  I manually verified that the uncovered lines still work.

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -203,17 +204,13 @@ def read(self, size: int = -1) -> bytes:
...
-            units, range_string = content_range.split(" ")
-            if units == "bytes":
-                range, total = range_string.split("/")
-                if "-" in range:
-                    _, end = range.split("-")
-                    end_pos = int(end)
-                    if total != "*" and end_pos >= int(total) - 1:
-                        self._eof = True
-            else:
-                self._log.warning("Requested byte range from server but instead got: %s", content_range)
```
dhirving, line 216: There's a slight behavior change here.  Previously if the units were anything but `bytes` it would only log a warning.  Now it throws an exception.  This should never happen in real life, and if it does the returned data is almost certainly wrong so we want to bail anyway.

python/lsst/resources/_resourceHandles/_httpResourceHandle.py
```diff
@@ -223,3 +220,60 @@ def read(self, size: int = -1) -> bytes:
...
+    ValueError
+        If the header was not in the expected format.
+    """
+    # There are three possible formats for Content-Range. All of them start
+    # with optional whitespace and a unit, which for our purposes should always
+    # be "bytes".
+    prefix = r"^\s*bytes\s+"
+
+    # Content-Range: <unit> <range-start>-<range-end>/<size>
+    if (case1 := re.match(prefix + r"(\d+)-(\d+)/(\d+)", header)) is not None:
```
timj, line 266: Hopefully three separate regexes won't turn into a hotspot.

timj (approved):
Looks great.

PR #78, opened 2024-01-19
DM-42568: Handle zero-byte files in S3 presigned URLs

dhirving:
Fix an issue for presigned S3 HTTP URLs with a zero byte length where HttpResourcePath.exists() would always return False, and HttpResourcePath.size() would throw an exception.

Because we emulate HEAD using a 1-byte GET request when checking S3 presigned HTTP URLs, the server returns 416 status instead of 206 because the range is longer than the file.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -823,11 +823,11 @@ class HttpResourcePath(ResourcePath):
                         f"Response to HEAD request to {self} does not contain 'Content-Length' header"
                     )
             elif resp.status_code == requests.codes.partial_content:
-                # 206, returned from a GET request with a Range header (used to
-                # emulate HEAD for presigned S3 URLs).  In this case
-                # Content-Length is the length of the Range and not the full
-                # length of the file, so we have to parse Content-Range
-                # instead.
+                # 206 Partial Content, returned from a GET request with a Range
+                # header (used to emulate HEAD for presigned S3 URLs).
+                # In this case Content-Length is the length of the Range and
+                # not the full length of the file, so we have to parse
+                # Content-Range instead.
                 content_range_header = resp.headers.get("Content-Range")
                 if content_range_header is None:
                     raise ValueError(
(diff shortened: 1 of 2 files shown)


python/lsst/resources/http.py
```diff
@@ -838,7 +838,16 @@ def size(self) -> int:
                 if size is None:
                     raise ValueError(f"Content-Range header for {self} did not include a total file size")
                 return size
-
+            elif resp.status_code == requests.codes.range_not_satisfiable:
+                # 416 Range Not Satisfiable, which can occur on a GET for a 0
+                # byte file since we asked for 1 byte Range which is longer
+                # than the file.
+                #
+                # Servers are supposed to include a Content-Range header in
```
timj, line 846: I will say in passing that we have in the past managed to get Google to fix their S3 server implementation.

What happens if you have a GCS signed URL? Same problem or does it really matter whether the URI was s3 or gs originally?

timj (approved):

python/lsst/resources/http.py
dhirving: Not sure -- I don't have a convenient way to generate signed GCS URLs.

The 416 thing isn't Google-specific and is standards-compliant behavior, btw, so this change isn't only necessary for Google compatibility.  I just failed to consider what would happen with 0-byte files when I did the initial implementation.

The part that is non-compliant is that it doesn't also return a Content-Range including the size (specified as a SHOULD in the HTTP RFC).  Unless something else really weird and non-compliant is happening that size will always be 0 so I don't think it's especially important.  

(If the server doesn't want to or can't deal with the Range at all it's supposed to return 200 with the full content instead, which will result in correct behavior since we go down the branch that looks at Content-Length.  Slightly rude to the server but probably mostly theoretical -- implementations of S3 will almost certainly support Range requests.)

It looks like the HttpResourceHandle also assumes it's at EOF and bails early without trying to look at Content-Range if it gets a 416, so I don't think it's a problem.

PR #80, opened 2024-01-23
DM-41547: Remove test dependencies from s3 and https deps

dhirving:
The s3 and https optional dependency groups included dependencies that are only needed for running tests.  This was causing downstream users to pull in unnecessary dependencies.  These deps have now been moved to the "test" group.

## Checklist

- [x] ran Jenkins


timj (approved):
It does raise the problem that if they ask for `test` they now get the s3 packages regardless. I imagine that's not really a problem because no-one is actually asking for `test` and having to do a `test+s3` seems overkill.

PR #81, opened 2024-01-29
DM-42698: Switch to the moto v5 API

timj:
## Checklist

- [x] ran Jenkins


ktlim (approved):

ktlim:
Except I guess I don't approve the test failure :)

timj:
Yes. You jumped the gun a bit 😄. The failure is a bit odd because it is acting like one batch of writes is being completely ignored by moto. Works fine in v4. 

Daf_butler needs fixing as well.

timj:
The tests are failing because only the first 1000 datasets from each loop seem to be written. If I change the counter to 900 in both loops then 1800 datasets are found. But if any of the loops cross 1000 it all fails. This is a bit odd since this is a simple loop running `write()`.

timj:
I am wondering if this is caused by https://github.com/getmoto/moto/issues/7152

timj:
The tests are failing because pagination in moto doesn't seem to be paginating. It's only returning the first 1000 results and no second page.

timj:
The moto problem is being fixed in getmoto/moto#7283

timj:
We could possibly change the constraint to `!=5.0.0` since the fix will appear in 5.0.1.

ktlim (approved):

PR #84, opened 2024-04-08
DM-43739: Fix memory leak in HTTPResourcePath

dhirving:
Fixed an issue where query parameters or fragments in an HTTP URL would cause it to be considered as having a different root URI.  This caused a new HTTP session to be created and cached each time an HTTP URL with a different query parameter was accessed.

This resulted in a memory leak in datalinker when it was using HTTPResourcePath to access S3 presigned URLs, which always have unique query parameters.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourcePath.py
+++ b/python/lsst/resources/_resourcePath.py
@@ -417,7 +417,7 @@ class ResourcePath:  # numpydoc ignore=PR02
         uri : `ResourcePath`
             Root URI.
         """
-        return self.replace(path="", forceDirectory=True)
+        return self.replace(path="", query="", fragment="", params="", forceDirectory=True)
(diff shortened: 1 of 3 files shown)


python/lsst/resources/_resourcePath.py
```diff
@@ -417,7 +417,7 @@ def root_uri(self) -> ResourcePath:
         uri : `ResourcePath`
             Root URI.
         """
-        return self.replace(path="", forceDirectory=True)
+        return self.replace(path="", query="", fragment="", forceDirectory=True)
```
timj, line 420: `params` as well?

timj (approved):
Thanks for finding this problem.

PR #85, opened 2024-05-20
DM-44398: Use temporary_uri inside transfer_from

timj:
This allows umask to be used when creating the files.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_file.py
+++ b/tests/test_file.py
@@ -148,12 +148,12 @@ class FileReadWriteTestCase(GenericReadWriteTestCase, unittest.TestCase):
         with ResourcePath.temporary_uri(suffix=".yaml", delete=False) as tmp:
             tmp.write(b"1234")
             self.assertTrue(tmp.exists(), f"uri: {tmp}")
-        exists = tmp.exists()
-        try:
-            tmp.remove()
-        except Exception:
-            pass
-        self.assertTrue(exists, f"uri: {tmp} should still exist")
+        # If the file doesn't exist there is nothing to clean up so a failure
+        # here is not a problem.
+        self.assertTrue(tmp.exists(), f"uri: {tmp} should still exist")
+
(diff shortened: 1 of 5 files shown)


timj:
@erykoff does this work for you?

erykoff:
It may work for me, but it doesn't work for mypy! 😂  I'll test this out after my mid-day meeting block.

timj:
Maybe @natelust can tell me if it would make sense for `ResourceHandleProtocol` to act like an `IO` when type checking.

tests/test_file.py
```diff
@@ -145,6 +145,16 @@ def test_local_temporary(self):
             self.assertTrue(tmp.exists())
         self.assertFalse(tmp.exists(), f"uri: {tmp}")
 
+        with ResourcePath.temporary_uri(suffix=".yaml", delete=False) as tmp:
+            tmp.write(b"1234")
+            self.assertTrue(tmp.exists(), f"uri: {tmp}")
+        exists = tmp.exists()
+        try:
+            tmp.remove()
+        except Exception:
```
dhirving, line 154: It might be better not to swallow all exceptions here... isn't part of the contract of a temporary file that it should be deletable?

Also I found it slightly confusing that the assertion for exists came after the call to `tmp.remove()` -- at first glance this test seemed to be stating that `tmp.remove()` is supposed to fail and the file should still exist at the end.

dhirving (approved):

tests/test_file.py
timj: Right. Thanks. I was in a rush and messed up. I realized that if the exists assert fails the clean up line wouldn't run so I did the check outside the assert, but of course, if the exist fails it means that the file isn't there to clean up anyhow so it's not a problem. I'll fix.

PR #86, opened 2024-05-22
DM-44486: Fix SEEK_END for HTTP and S3 handles

timj:
## Checklist

- [x] ran Jenkins


python/lsst/resources/_resourceHandles/_s3ResourceHandle.py
```diff
@@ -239,8 +257,7 @@ def seek(self, offset: int, whence: int = SEEK_SET) -> int:
             elif whence == SEEK_CUR:
                 self._position += offset
             elif whence == SEEK_END:
-                offset = abs(offset)
```
ktlim, line 242: So this was actually broken before?

timj: Completely...

PR #91, opened 2024-08-09
DM-44547: Add fsspec constructor

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourcePath.py
+++ b/python/lsst/resources/_resourcePath.py
@@ -11,6 +11,8 @@
 
 from __future__ import annotations
 
+__all__ = ("ResourcePath", "ResourcePathExpression")
+
 import concurrent.futures
 import contextlib
 import copy
@@ -26,7 +28,12 @@ import urllib.parse
 from pathlib import Path, PurePath, PurePosixPath
 from random import Random
 
-__all__ = ("ResourcePath", "ResourcePathExpression")
+try:
+    import fsspec
+    from fsspec.spec import AbstractFileSystem
+except ImportError:
+    fsspec = None
+    AbstractFileSystem = type
 
 from collections.abc import Iterable, Iterator
 from typing import TYPE_CHECKING, Any, Literal, overload
@@ -408,6 +415,21 @@ class ResourcePath:  # numpydoc ignore=PR02
             String form of URI.
         """
         return self._uri.geturl()
+
+    def to_fsspec(self) -> tuple[AbstractFileSystem, str]:
+        """Return an abstract file system and path that can be used by fsspec.
+
+        Returns
+        -------
+        fs : `fsspec.spec.AbstractFileSystem`
+            A file system object suitable for use with the returned path.
+        path : `str`
+            A path that can be opened by the file system object.
+        """
+        if fsspec is None:
+            raise ImportError("fsspec is not available")
+        # By default give the URL to fsspec and hope.
+        return fsspec.url_to_fs(self.geturl())
 
     def root_uri(self) -> ResourcePath:
         """Return the base root URI.
--- a/python/lsst/resources/s3.py
+++ b/python/lsst/resources/s3.py
@@ -32,6 +32,8 @@ from ._resourceHandles._s3ResourceHandle import S3ResourceHandle
 from ._resourcePath import ResourcePath
 from .s3utils import (
     _get_s3_connection_parameters,
+    _s3_disable_bucket_validation,
+    _s3_should_validate_bucket,
     all_retryable_errors,
     backoff,
     bucketExists,
@@ -335,6 +337,11 @@ class S3ResourcePath(ResourcePath):
             key=endpoint_config.access_key_id,
             secret=endpoint_config.secret_access_key,
         )
+        if not _s3_should_validate_bucket():
+            # Accessing the s3 property forces the boto client to be
+            # constructed and cached and allows the validation to be removed.
+            _s3_disable_bucket_validation(s3.s3)
+
         return s3, f"{self._bucket}/{self.relativeToPathRoot}"
 
     def _as_local(self) -> tuple[str, bool]:
--- a/python/lsst/resources/s3.py
+++ b/python/lsst/resources/s3.py
@@ -31,6 +31,7 @@ from ._resourceHandles._baseResourceHandle import ResourceHandleProtocol
 from ._resourceHandles._s3ResourceHandle import S3ResourceHandle
 from ._resourcePath import ResourcePath
 from .s3utils import (
+    _get_s3_connection_parameters,
     all_retryable_errors,
     backoff,
     bucketExists,
@@ -45,6 +46,13 @@ try:
     from boto3.s3.transfer import TransferConfig  # type: ignore
 except ImportError:
     TransferConfig = None
+
+try:
+    import fsspec
+    from fsspec.spec import AbstractFileSystem
+except ImportError:
+    fsspec = None
+    AbstractFileSystem = type
 
 if TYPE_CHECKING:
     with contextlib.suppress(ImportError):
@@ -306,6 +314,28 @@ class S3ResourcePath(ResourcePath):
         except ClientError as err:
             translate_client_error(err, self)
             raise
+
+    def to_fsspec(self) -> tuple[AbstractFileSystem, str]:
+        """Return an abstract file system and path that can be used by fsspec.
+
+        Returns
+        -------
+        fs : `fsspec.spec.AbstractFileSystem`
+            A file system object suitable for use with the returned path.
+        path : `str`
+            A path that can be opened by the file system object.
+        """
+        if fsspec is None:
+            raise ImportError("fsspec is not available")
+        # Must remove the profile from the URL and form it again.
+        endpoint_config = _get_s3_connection_parameters(self._profile)
+        return fsspec.url_to_fs(
+            f"s3://{self._bucket}/{self.relativeToPathRoot}",
+            profile=endpoint_config.profile,
+            endpoint_url=endpoint_config.endpoint_url,
+            key=endpoint_config.access_key_id,
+            secret=endpoint_config.secret_access_key,
+        )
 
     def _as_local(self) -> tuple[str, bool]:
         """Download object from S3 and place in temporary directory.
--- a/python/lsst/resources/s3.py
+++ b/python/lsst/resources/s3.py
@@ -48,10 +48,10 @@ except ImportError:
     TransferConfig = None
 
 try:
-    import fsspec
+    import s3fs
     from fsspec.spec import AbstractFileSystem
 except ImportError:
-    fsspec = None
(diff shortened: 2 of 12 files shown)


timj:
@erykoff from looking at fsspec it seems like they want me to do the equivalent of:

```python
import fsspec

fs, path = fs_to_url(ResourcePath(uri).geturl())
with fs.open(path) as fd:
    content = fd.read()
```
which is what I've implemented here. That does "just work" for file and https. It won't work for webdav and s3 where the credentials need to be given to `fs_to_url` so I think a `ResourcePath.to_fsspec()` is still needed for to call `fs_to_url` behind the scenes with the correct parameters before the formatter does the `fs.open()` itself. The alternative is I implement something like:

```python
r = ResourcePath(uri)
with r.open_as_fsspec("rb") as fd:
    content = fd.read()
```

erykoff:
To be clear, what I’m looking for is “an fsspec subclass” whatever that means https://arrow.apache.org/docs/python/filesystems.html#using-fsspec-compatible-filesystems-with-arrow

timj:
That looks to be exactly what url to fs gives you so I think it's fine. You can try it on this branch for https and file locations.

tests/test_s3.py
```diff
@@ -225,6 +231,17 @@ def test_threading_false(self):
...
+    @unittest.skipIf(fsspec is None, "fsspec is not available")
+    def test_fsspec_constructor(self) -> None:
+        """Test that we can obtain an s3fs object."""
+        uri = self.root_uri.join("test_file.dat")
+        fs, path = uri.to_fsspec()
+        self.assertEqual(path, f"{uri._bucket}/{uri.relativeToPathRoot}")
+        self.assertTrue(hasattr(fs, "open"))
+
+    def test_fsspec(self) -> None:
+        raise unittest.SkipTest("fsspec s3fs incompatible with moto")
```
dhirving, line 243: Using moto in [server mode](https://docs.getmoto.org/en/stable/docs/server_mode.html#start-within-python) would be one way to get this test working.  Might not be worth the trouble though.

python/lsst/resources/_resourcePath.py
```diff
@@ -409,6 +416,21 @@ def geturl(self) -> str:
...
+        """Return an abstract file system and path that can be used by fsspec.
+
+        Returns
+        -------
+        fs : `fsspec.spec.AbstractFileSystem`
+            A file system object suitable for use with the returned path.
+        path : `str`
+            A path that can be opened by the file system object.
+        """
+        if fsspec is type:
```
dhirving, line 429: `fsspec` is `None` when it's not available, so this exception will never raise.  (This same issue is present in all the files.)

python/lsst/resources/s3utils.py
```diff
@@ -230,26 +251,17 @@ def getS3Client(profile: str | None = None) -> boto3.client:
...
-    skip_validation = not re.search(r"^(0|f|n|false)?$", disable_value, re.I)
-
-    return _get_s3_client(endpoint, profile, skip_validation)
+    return _parse_endpoint_config(endpoint, profile)
 
 
 @functools.lru_cache
-def _get_s3_client(endpoint: str | None, profile: str | None, skip_validation: bool) -> boto3.client:
+def _get_s3_client(
+    endpoint_config: _EndpointConfig, profile: str | None, skip_validation: bool
```
dhirving, line 259: This `profile` parameter is unused.

python/lsst/resources/s3utils.py
```diff
@@ -230,26 +251,17 @@ def getS3Client(profile: str | None = None) -> boto3.client:
...
-    endpoint_config = _parse_endpoint_config(endpoint)
-
-    if endpoint_config.access_key_id is not None and endpoint_config.secret_access_key is not None:
-        # We already have the necessary configuration for the profile, so do
-        # not pass the profile to boto3.  boto3 will raise an exception if the
-        # profile is not defined in its configuration file, whether or not it
-        # needs to read the configuration from it.
-        profile = None
-    session = boto3.Session(profile_name=profile)
+    session = boto3.Session(profile_name=endpoint_config.profile)
```
dhirving, line 265: A few lines down (line 273) we run this code to disable bucket name validation in botocore.
```
 if skip_validation:
    client.meta.events.unregister("before-parameter-build.s3", validate_bucket_name)
```

This is apparently needed to make Ceph multi-tenant bucket names work.  Are we still using Ceph multi-tenant buckets?  It looks like fsspec uses s3fs which uses botocore internally, so we may need to figure out how to apply a similar workaround when using fsspec.

dhirving (approved):
Looks good to me.

We will eventually need WebDAV support if anything Eli is using this for will be involved in the data processing at FrDF though.

python/lsst/resources/_resourcePath.py
timj: Thanks -- that was from earlier version. The perils of not testing without it installed.

dhirving:
It looks like one of the tests fails without fsspec installed as well:
```
__________________________________________ ResourceReadTestCase.test_read ___________________________________________

self = <tests.test_resource.ResourceReadTestCase testMethod=test_read>

    def test_read(self):
        uri = self.root_uri.join("Icons/README.txt")
        self.assertTrue(uri.exists(), f"Check {uri} exists")
    
        content = uri.read().decode()
        self.assertIn("IDLE", content)
    
        with uri.as_local() as local_uri:
            self.assertEqual(local_uri.scheme, "file")
            self.assertTrue(local_uri.exists())
    
        truncated = uri.read(size=9).decode()
        self.assertEqual(truncated, content[:9])
    
        # Check that directory determination can work directly without the
        # trailing slash.
        d = self.root_uri.join("Icons")
        self.assertTrue(d.isdir())
        self.assertTrue(d.dirLike)
    
        d = self.root_uri.join("Icons/", forceDirectory=True)
        self.assertTrue(uri.exists(), f"Check directory {d} exists")
        self.assertTrue(d.isdir())
    
        with self.assertRaises(IsADirectoryError):
            with d.as_local() as local_uri:
                pass
    
        j = d.join("README.txt")
        self.assertEqual(uri, j)
        self.assertFalse(j.dirLike)
        self.assertFalse(j.isdir())
        not_there = d.join("not-there.yaml")
        self.assertFalse(not_there.exists())
    
        bad = ResourcePath(f"{self.scheme}://bad.module/not.yaml")
        multi = ResourcePath.mexists([uri, bad, not_there])
        self.assertTrue(multi[uri])
        self.assertFalse(multi[bad])
        self.assertFalse(multi[not_there])
    
        # Check that the bad URI works as expected.
        self.assertFalse(bad.exists())
        self.assertFalse(bad.isdir())
            bad.read()
            with bad.as_local():
                pass
            with bad.open("r"):
                pass
    
        # fsspec is always not implemented.
        with self.assertRaises(NotImplementedError):
>           bad.to_fsspec()
E           AttributeError: 'PackageResourcePath' object has no attribute 'to_fsspec'

tests/test_resource.py:96: AttributeError
```

timj:
> It looks like one of the tests fails without fsspec installed as well

Thanks. Will take a look. It's odd that the entire method is missing.

timj:
> 'PackageResourcePath' object has no attribute 'to_fsspec'

I can't reproduce this. It's what would happen if you weren't using the ticket branch.

tests/test_s3.py
timj: Yes. I found a couple of posts online about changing to server mode and decided that was more work reorganizing all the tests than seemed warranted for the moment.

python/lsst/resources/s3utils.py
timj: @ktlim can you remind me of an example place at USDF where we use the names that don't work so I can try this out?

dhirving:
> > 'PackageResourcePath' object has no attribute 'to_fsspec'
> 
> I can't reproduce this. It's what would happen if you weren't using the ticket branch.

Sorry, I forgot to `setup -k -r` in my local checkout of the resources repo.  So it was in fact not using the ticket branch.

python/lsst/resources/s3utils.py
timj: @erykoff confirms that a URI such as `"s3://rubin:rubinobs-lfa-cp/Electrometer:101/fits/2024/07/29/Electrometer:101_fits_2024-07-29T17:26:16.810.fits"` fails because fsspec parses the URL and thinks it has a port number. We might have to create an explicit s3fs object instead of using `url_to_fs`.

python/lsst/resources/s3.py
```diff
@@ -307,6 +317,32 @@ def _download_file(
...
+        endpoint_config = _get_s3_connection_parameters(self._profile)
+        s3 = s3fs.S3FileSystem(
+            profile=endpoint_config.profile,
+            endpoint_url=endpoint_config.endpoint_url,
+            key=endpoint_config.access_key_id,
+            secret=endpoint_config.secret_access_key,
+        )
+        if not _s3_validate_bucket():
+            # This uses a private property for now.
+            s3._s3.meta.events.unregister("before-parameter-build.s3", validate_bucket_name)
```
timj, line 342: This does not work because the boto client is not constructed until it's used. We would have to force the S3FileSystem object to do something before returning it.

timj:
@dhirving can you please take another look at the s3 implementation? I ended up having to use s3fs directly because the bucket validation issue and then have to hack the internals to actually disable.

python/lsst/resources/s3.py
```diff
@@ -307,6 +317,35 @@ def _download_file(
...
+            endpoint_url=endpoint_config.endpoint_url,
+            key=endpoint_config.access_key_id,
+            secret=endpoint_config.secret_access_key,
+        )
+        if not _s3_validate_bucket():
+            # Force the boto client to be constructed so we can modify it.
+            with contextlib.suppress(FileNotFoundError):
+                s3.open("bad_bucket/path")
+            # This uses a private property for now.
+            s3._s3.meta.events.unregister("before-parameter-build.s3", validate_bucket_name)
```
dhirving, line 345: `S3FileSystem` has a public `s3` property you can use to create and get the boto client.

```suggestion
            s3.s3.meta.events.unregister("before-parameter-build.s3", validate_bucket_name)
```

dhirving, line 345: It might be a good idea to roll the events-unregistering hack into a little utility function so it's de-duplicated between the two files, and maybe add a comment explaining what it is doing since it's otherwise a bit mysterious.

dhirving (approved):

python/lsst/resources/s3.py
timj: Thanks. I feel a bit better about a public property.

dhirving:
Also I wonder if the s3fs object wants to be cached.  When we instantiate an s3 client ourselves, we cache it (though whether that is necessary or not I'm not sure.)  S3FS is doing the same thing internally, so if it's necessary in the other place it might be necessary here.

PR #92, opened 2024-08-12
DM-45732: Catch NoSuchKey in s3 handle read

timj:
## Checklist

- [ ] ran Jenkins


erykoff (approved):
I guess this is fine? I'm not really suited to these sorts of things.

timj:
Does it work for you?

PR #94, opened 2024-09-26
DM-45978: Add webdav support to ResourcePath.to_fsspec

airnandez:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -376,8 +376,9 @@ class HttpResourcePathConfig:
 
 
 @functools.lru_cache
-def _is_webdav_endpoint(path: ResourcePath | str) -> bool:
-    """Check whether the remote HTTP endpoint implements WebDAV features.
+def _get_dav_and_server_headers(path: ResourcePath | str) -> tuple[str | None, str | None]:
+    """Retrieve the "DAV" and "Server" headers sent by the remote server as
+    part of the response to a single "OPTIONS" HTTP request.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


python/lsst/resources/http.py
```diff
@@ -34,8 +37,16 @@
...
+try:
+    import fsspec
+    from fsspec.spec import AbstractFileSystem
+except ImportError:
+    fsspec = None
+    AbstractFileSystem = type
+
 from urllib.parse import parse_qs
 
+import aiohttp
```
timj, line 49: I think this will likely need an RFC to formally add it to rubinenv conda environment. I think it comes in through webdav4 at the moment so things will work. Also need to list it in requirements.txt and pyproject.toml (the latter for the http option) -- or else put protections around the import if this is only used with fsspec.

timj: (regardless of import protections we still formally need rubinenv to list it since we want to ensure that this code works in production).

airnandez: I think `aiohttp` is already in the environment, since I could find it. In fact, `fsspec` is built on top of it.

airnandez:
```bash
$ source /cvmfs/sw.lsst.eu/almalinux-x86_64/lsst_distrib/w_2024_38/loadLSST.sh

$ command -v python
/cvmfs/sw.lsst.eu/almalinux-x86_64/lsst_distrib/w_2024_38/conda/envs/lsst-scipipe-9.0.0-exact/bin/python

$ python -c "import aiohttp"

$ echo $?
0
```

airnandez: To be precise, I am _not_ using `webdav4`. I preferred not to since after 4 weeks I got no response to an issue I submitted to them, so I don't want to depend on that package (see https://github.com/skshetry/webdav4/issues/196)

python/lsst/resources/http.py
```diff
@@ -1137,6 +1218,116 @@ def walk(
...
+            raise ImportError("fsspec is not available")
+
+        async def get_client(**kwargs: Any) -> aiohttp.ClientSession:
+            # Create a SSL context equipped with the required certificates of
+            # the trusted certificate authorities. fsspect client must use
+            # that context to download the file.
+            ssl_context = ssl.create_default_context()
+            if self._config.ca_bundle is not None:
+                ssl_context.load_verify_locations(capath=self._config.ca_bundle)
+            return aiohttp.ClientSession(connector=aiohttp.TCPConnector(ssl=ssl_context), **kwargs)
```
dhirving, line 1278: I wonder if the `TCPConnector` should be cached to allow connection pooling, similar to what is done with the `requests` `Session` objects.

python/lsst/resources/http.py
```diff
@@ -1137,6 +1218,116 @@ def walk(
...
+
+        # Retrieve a signed URL for a duration of 1 hour. We assume the URL
+        # is to be used only for downloading, to avoid the risk the caller
+        # modify its contents inadvertently.
+        url = self.generate_presigned_get_url(expiration_time_seconds=3_600)
```
dhirving, line 1283: This will break `to_fsspec` for non-webdav HTTP/HTTPS URLs, since fetching a macaroon will fail.  Previously it was inheriting the base implementation from `ResourcePath.to_fsspec`, which just passes the URL unchanged to `fsspec.url_to_fs`.

Maybe add a conditional at the top delegating to the base class for non-webdav URLs?

python/lsst/resources/http.py
```diff
@@ -1137,6 +1218,116 @@ def walk(
...
+        try:
+            response_body = json.loads(resp.text)
+            if "macaroon" in response_body:
+                return f"{self}?authz={response_body['macaroon']}"
+            else:
+                raise ValueError(f"could not retrieve macaroon for URL {self}")
+        except json.JSONDecodeError:
+            raise ValueError(f"could not deserialize response to POST request for URL {self}")
+        except Exception as e:
+            raise e
```
dhirving, line 1329: Since it just re-raises, the behavior would be the same here without the explicit `except Exception`.

python/lsst/resources/http.py
dhirving: I agree with Fabio -- I don't think the RFC is needed.  We only need `aiohttp` if we have `fsspec`, and having `fsspec` implies `aiohttp`.  The interface Fabio is using requires him to pass an `aiohttp` client to `fsspec`, so if `fsspec` stops using `aiohttp` internally this is already going to break.

However, I think this import should be part of the `try` block for `fsspec`, or lazy-import it in the function below where it is used.  We have some services that install this library via `pip`.  `fsspec` isn't listed as a package dependency so we don't get it by default.

python/lsst/resources/http.py
```diff
@@ -1137,6 +1218,116 @@ def walk(
...
+        if self.server not in ("dcache", "xrootd"):
+            raise NotImplementedError(f"server '{self._server}' does not support signing for {self}")
+
+        match activity:
+            case ActivityCaveat.DOWNLOAD:
+                activity_caveat = "DOWNLOAD,LIST"
+            case ActivityCaveat.UPLOAD:
+                activity_caveat = "UPLOAD,LIST"
+
+        # Request a macaroon for the requested activities and duration duration
```
dhirving, line 1307: It might be helpful to include a link to some documentation explaining this macaroon stuff...

I eventually found https://www.dcache.org/manuals/workshop-2017-05-29-Umea/000-Final/anupam_macaroons_v02.pdf which helped me figure out what was going on but it took a minute.

dhirving (approved):
Looks good to me other than that it breaks `to_fsspec` for non-webdav HTTP URLs.  Should be an easy fix though.

Really good to know that `dcache` and `xrootd` support a way to generate pre-signed URLs.  That means we can use them as a backend for the Butler server, which might be useful at some point.

python/lsst/resources/http.py
dhirving: It might also be good to get the call to `load_verify_locations` out of the async function, since it does blocking I/O.  Probably fast enough it's not a huge deal though.

python/lsst/resources/http.py
airnandez: ✅ Done: import guarded by same `try` block used for `fsspec`

python/lsst/resources/http.py
airnandez: ✅ Done: `TCPconnector` cached and `load_verify_locations` called out of the async function.

python/lsst/resources/http.py
airnandez: ✅ Done: conditional added. `ResourcePath.to_fsspec` called for plain HTTP paths.

python/lsst/resources/http.py
airnandez: ✅ Done

python/lsst/resources/http.py
airnandez: ✅ Done: pointer to official documentation added.

python/lsst/resources/http.py
timj: What I'm trying to say is that if we explicitly need aiohttp for this code to work the way we want (ie fsspec working on webdav in the production environment) then we need to make sure that rubin-env comes with aiohttp even if internally fsspec drops that dependency.

dhirving: Tim -- this code is guaranteed to not work if fsspec drops that dependency in any case.  The fsspec constructor signature we are calling is:
```
HTTPFileSystem(get_client: aiohttp.ClientSession)
``` 

So if fsspec stops using aiohttp internally, it's going to break no matter what we do with the conda environment.

PR #95, opened 2024-09-30
DM-46139: Add option to disable multithreaded S3 uploads

andy-slac:
We already have an option to disable multithreading in downloads via `LSST_S3_USE_THREADS` environment variable. This option may be useful for uploads too, updating the code to use it for uploads.

## Checklist

- [x] ran Jenkins
- [X] added a release note for user-visible changes to `doc/changes`


andy-slac:
Python 3.13 fails, looks like `Icons/README.txt` file was updated in Python distribution.

timj (approved):

PR #96, opened 2024-10-01
DM-42091: Remove long deprecated s3 utilities

timj:
## Checklist

- [x] ran Jenkins


timj:
The python 3.13 failure is going to fix itself because 3.13-rc3 has the text added back that was removed in rc2 (and was present in rc1).

dhirving (approved):

PR #98, opened 2024-10-18
DM-46776: Propagate fragments with ResourcePath.join

timj:
Previously fragments were either dropped completely (if a ResourcePath was given that had a fragment) or included in the file name (if a plain string was given).

Now fragments are always propagated if the ResourcePath parsing finds a fragment. This means that if a filename has a "#" in it that "#" will now always be treated as a fragment and not part of the filename.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/_resourcePath.py
+++ b/python/lsst/resources/_resourcePath.py
@@ -693,7 +693,7 @@ class ResourcePath:  # numpydoc ignore=PR02
             quoted depending on the associated URI scheme. If the path looks
             like a URI referring to an absolute location, it will be returned
             directly (matching the behavior of `os.path.join`). It can
-            also be a `ResourcePath`.
+            also be a `ResourcePath`. Fragments are propagated.
         isTemporary : `bool`, optional
             Indicate that the resulting URI represents a temporary resource.
             Default is ``self.isTemporary``.
@@ -722,6 +722,9 @@ class ResourcePath:  # numpydoc ignore=PR02
         It is an error to attempt to join to something that is known to
         refer to a file. Use `updatedFile` if the file is to be
         replaced.
+
+        If an unquoted ``#`` is included in the path it is assumed to be
+        referring to a fragment and not part of the file name.
 
         Raises
         ------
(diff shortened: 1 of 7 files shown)


python/lsst/resources/_resourcePath.py
```diff
@@ -757,10 +765,10 @@ def join(
...
             return path_uri
 
-        # If this was originally a ResourcePath extract the unquoted path from
-        # it. Otherwise we use the string we were given to allow "#" to appear
-        # in the filename if given as a plain string.
-        if not isinstance(path, str):
+        # We want to propagate fragments to the joined path and we rely on
+        # the ResourcePath parser to find these fragments for us even in plain
+        # strings. Must assume there are no `#` characters in filenames.
+        if not isinstance(path, str) or path_uri.fragment:
```
dhirving, line 771: I'm a little sketched out by this -- it seems surprising that if you give `join()` a string filename it doesn't use that verbatim as the filename.  It works within the specific context of how the datastore is set up, but breaks more general usage.  I also wonder if there are some (probably slight) security implications here, where someone  could smuggle in a `..#something` or similar

It seems to me that callers could use `.replace(fragment=...)` if they want to add a fragment, or explicitly convert the string to a `ResourcePath` before calling `join()` if their intention is to treat it as a URI and not a path fragment.

dhirving (approved):
Looks OK to me.

python/lsst/resources/_resourcePath.py
timj: I think the difficulty here is that before this change:
```python
>>> from lsst.resources import ResourcePath
>>> ResourcePath("abc#def.txt")._uri
ParseResult(scheme='file', netloc='', path='/Users/timj/work/lsstsw/build/resources/abc', params='', query='', fragment='def.txt')
>>> ResourcePath("").join("abc#def.txt")._uri
ParseResult(scheme='file', netloc='', path='/Users/timj/work/lsstsw/build/resources/abc%23def.txt', params='', query='', fragment='')
```

and with this change:

```
>>> ResourcePath("abc#def.txt")._uri
ParseResult(scheme='file', netloc='', path='/Users/timj/work/lsstsw/build/resources/abc', params='', query='', fragment='def.txt')
>>> ResourcePath("").join("abc#def.txt")._uri
ParseResult(scheme='file', netloc='', path='/Users/timj/work/lsstsw/build/resources/abc', params='', query='', fragment='def.txt')
```
You get a consistent answer.  These relative file paths can come from users specifying subsets of pipelines or from relative path entries in datastore.

We are assuming in other places that a `#` in the file part of a schemeless URI means fragment and that no-one using this is going to be putting a `#` in the filename itself. The question is whether we break that assumption elsewhere or try to make that assumption consistent.

timj: For completeness, I think there is one other potential problem with "# in filename" at the moment.

```
>>> from lsst.resources import ResourcePath
>>> x = ResourcePath("a/bc#de.txt", forceAbsolute=False)
>>> y = ResourcePath("a", forceAbsolute=False)
>>> str(x)
'a/bc#de.txt'
>>> str(y.join("bc#de.txt"))
'a/bc#de.txt'
```
And for before and after this branch they stringify the same even though before one of those was a filename in the internal _uri representation and the other was a fragment.

PR #99, opened 2024-11-05
DM-47357: Explicitly configure timeouts for fsspec

airnandez:
The default timeout and number of TCP connections kept alive by fsspec' HTTPFileSystem are not suitable for production conditions so we set them to more reasonable values based on observed behavior.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/resources/http.py
+++ b/python/lsst/resources/http.py
@@ -125,6 +125,7 @@ class HttpResourcePathConfig:
         self._back_end_connections: int | None = None
         self._digest_algorithm: str | None = None
         self._send_expect_on_put: bool | None = None
+        self._fsspec_is_enabled: bool | None = None
         self._timeout: tuple[float, float] | None = None
         self._collect_memory_usage: bool | None = None
         self._backoff_min: float | None = None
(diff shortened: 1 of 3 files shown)


python/lsst/resources/http.py
```diff
@@ -1325,14 +1325,43 @@ async def get_client_session(**kwargs: Any) -> ClientSession:
...
+                )
 
-            return ClientSession(connector=HttpResourcePath._tcp_connector, **kwargs)
+            connect_timeout, read_timeout = self._config.timeout
+            return ClientSession(
+                connector=HttpResourcePath._tcp_connector,
+                timeout=ClientTimeout(
+                    connect=connect_timeout,
+                    sock_connect=connect_timeout,
+                    sock_read=read_timeout,
```
dhirving, line 1355: This seems a bit long, potentially leaving a process stuck waiting for an actually-broken HTTP connection for 25 minutes.

But since you've been using this same read timeout for a long time with `requests`, it must not be an issue in practice.

python/lsst/resources/http.py
```diff
@@ -1325,14 +1325,43 @@ async def get_client_session(**kwargs: Any) -> ClientSession:
...
+                    # Total number of simultaneous connections this connector
+                    # keeps open with any host.
+                    #
+                    # The default is 100 but we deliberately reduced it to
+                    # avoid keeping a large number of open connexions to file
+                    # servers when thousands of quanta execute simultaneously.
+                    #
+                    # In any case, new connexions are automatically established
+                    # when needed.
+                    limit=10,
```
dhirving, line 1341: Out of curiosity, are the files spread across multiple HTTP endpoints with different hostnames at FrDF?

dhirving (approved):
Looks good to me.

python/lsst/resources/http.py
airnandez: (Sorry for not responding, I had missed this message)

Yes, there are several front-end hosts behind a DNS load-balancer. In addition, those front-end hosts redirect to the back end server where the requested file is physically located for that server to ultimately serve the request. In our current configuration there are several dozens of back end servers.

python/lsst/resources/http.py
airnandez: Yes, we have not seen issues with this in practice. We had to increase this timeout to cope with the non-understood issues related to our usage of `fsspec`. We may decrease that read timeout when we understand the root cause of the issue.

PR #107, opened 2025-03-04
DM-31824: Add experimental mtransfer class method

timj:
## Checklist

- [ ] ran Jenkins


python/lsst/resources/_resourcePath.py
```diff
@@ -896,6 +930,60 @@ def _mexists(cls, uris: Iterable[ResourcePath]) -> dict[ResourcePath, bool]:
...
+            A sequence of the source URIs and the target URIs.
+        overwrite : `bool`, optional
+            Allow an existing file to be overwritten. Defaults to `False`.
+        transaction : `~lsst.resources.utils.TransactionProtocol`, optional
+            A transaction object that can (depending on implementation)
+            rollback transfers on error.  Not guaranteed to be implemented.
+
+        Returns
+        -------
+        copy_status : `dict` [ `ResourcePath`, `bool` ]
```
timj, line 959: @dhirving once sticking point for `mtransfer` over `transfer_from` in a loop is that the latter raises an exception and tells you the problem immediately and you stop doing the transfers. `mtransfer` as written lets everything complete and then tells you a simple yes/no when the caller would like to know some reasons for failures (such as FileExistsError). I could store the Exceptions in the returned dict. Or I could let the first failure raise (is that allowed in concurrent futures?)? Or at the end of the transfers all the number of failures could be counted and the final exception encountered could be raised (with a note saying how many other failures there were). Should there be a parameter to indicate whether the caller wants a raise vs dict returned? I would be interested in your opinion on this. (butler does need to have some idea as to why a failure happened to try to help in its error message)

dhirving (approved):
I share Fabio's concern about the threadsafety of the HttpResourcePath implementation... even if the underlying library is threadsafe, it looks like our initialization logic might not be.  At least it looks like we could accidentally create multiple connection pools the first time `mtransfer` is called.  And there may be some other problematic bits in there since there is a lot of class-level caching.

PR #110, opened 2025-03-31
DM-49786: Add dav[s] scheme stubs

timj:
Currently become http internally.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-49786.misc.rst
+++ b/doc/changes/DM-49786.misc.rst
@@ -1,0 +1,1 @@
+Added support for ``dav`` URI schemes as aliases for HTTP.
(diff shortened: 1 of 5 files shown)


dhirving (approved):
I think this sort of works in that it allows the files to be read.  But when it comes time to write the URI into the database, it will convert into `http` instead of `dav`.  I wonder if it would be better to preserve it as `dav`/`davs`, and change it to `http` internally in `HttpResourcePath` when it goes to do an actual request.

timj:
Sounds reasonable. Will see if it all happens in one place.

timj:
@dhirving please take another look. There were quite a few changes.

dhirving (approved):
Looks good to me

PR #112, opened 2025-04-07
DM-49913: No longer use temporary directory for temp files

timj:
* Temp directory creation inside a temporary directory meant it was impossible to delete the temp directory if the temp file was not to be deleted. It's not clear we gained anything by having a temp dir with one tempfile in it.
* Change temporary_uri to use the same logic for determination of the temp directory as used in http.py

## Checklist

- [ ] ran Jenkins


jchiang87 (approved):
lgtm.

PR #124, opened 2025-06-13
DM-51384: Fix AccessDenied when copying between S3 with different profiles

dhirving:
Fix an issue where S3ResourcePath.transfer_from(S3ResourcePath) would fail when the source and destination were using different S3 endpoints or sets of credentials.  We now intermediate the download ourselves, since boto3 does not support transfers between multiple endpoints.

## Checklist

- [ ] ran Jenkins


timj (approved):
Thanks for fixing the error message.

ktlim (approved):
LGTM.  Thanks!

dhirving:
The Jenkins failure is unrelated fallout from a pytest dependency upgrade -- it passes on the other two environments.

PR #129, opened 2025-06-25
DM-51540: Avoid race condition in no-overwrite copies.

ktlim:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/resources/file.py
+++ b/python/lsst/resources/file.py
@@ -21,6 +21,7 @@ import os.path
 import posixpath
 import re
 import shutil
+import stat
 import urllib.parse
 from collections.abc import Iterator
 from typing import IO, TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


python/lsst/resources/file.py
```diff
@@ -328,15 +329,35 @@ def transfer_from(
...
+                        with open(temp_path, "x"):
+                            pass
+                    except FileExistsError:
+                        raise FileExistsError(
+                            f"Another process is writing to '{self}'."
+                            f" Transfer from {src} cannot be completed."
+                        )
+                    with transaction.undoWith(f"copy from {local_src}", os.remove, temp_path):
+                        st = os.stat(temp_path)
+                        os.chmod(temp_path, st.st_mode | stat.S_IWUSR)
```
timj, line 351: Maybe add a comment here that you re correcting for the restrictive mode that temp files are usually given?

timj (approved):
Looks great. Thanks.

PR #131, opened 2025-08-01
DM-52021: Include extra HTTP headers when creating fsspec

dhirving:
Attach "extra" authorization headers provided by the user to the fsspec instance created by `HttpResourcePath.to_fsspec()`.  This is needed to support the Butler setup at the CADC IDAC, which uses an HTTP datastore that requires authentication headers.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -104,6 +104,17 @@ class GenericHttpTestCase(GenericTestCase, unittest.TestCase):
         )
         self.assertEqual(path.read(), b"test")
 
+        # Make sure that headers are added to fsspec.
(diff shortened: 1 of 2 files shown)


tests/test_http.py
```diff
@@ -104,6 +104,17 @@ def test_extra_headers(self):
         )
         self.assertEqual(path.read(), b"test")
 
+        # Make sure that headers are added to fsspec.
+        # This triggers logic for "webdav" vs "not-webdav" that does an OPTIONS
```
timj, line 108: At some point we need to strip out the webdav stuff...

timj (approved):

